PR #47, opened 2019-02-21
ParsedJson & ParsedJson::iterator definitions in .cpp files

atomheartother:
This PR mostly intends to make the ParsedJson namespace's code more readable by moving stuff around without affecting performance.

# Changes
## Moved non-inline code to `.cpp` files
Having most code definitions in `.cpp` files and not in `.h` files makes the classes more readable, it also means less compile time for people using the library. Not to mention that this way people can upgrade to later versions without having to recompile their own parts of the project if we change code in these functions, and we gain some compiling speed. There are no performance downsides, so there's just no reason not to do it beyond wanting to be in line with the inline functions (no pun intended) and I personally don't think that's enough a reason. Inline functions are the exception, not the rule.

## Removed some non-critical inline functions
This change is debatable but I'd like to start that discussion. Not only are header files really hard to read due to all those inline functions, having too many inline functions can negatively affect binary size, and this is even more of a concern for a library. Of course, for performance reasons, we need *some* inline functions.

I first removed every inline function, which resulted in a drop of 0.4GB/s on parsing the test `twitter.json`. I then re-added functions one by one, seeing which functions were critical to the parsing speed.

You'll note no functions are inline in the iterator now. Since we don't have a precise benchmarking tool for ParsedJson::iterator, I haven't gone through that process quite as precisely, however `json2json` (with standard output redirected to `/dev/null`)didn't seem to behave any slower on the `twitter.json` file after I removed every inline function from it. (Timed very simply to the ms with my shell's `time` command)

## Minor changes
- I moved `ParsedJson::iterator::get_string()` in the header file to be next to the other `get_x` functions. Just seemed more sensible.
- Merged two CMakeList instructions doing the same thing, seemed a little silly to separate them
- I also took the liberty of adding `build/` to the `.gitignore` file (and of creating said `.gitignore` file)

# Notes
This branch was only tested on Linux, however it seems to be just as performant as origin, currently. It could affect backward-compatibility with the inlne functions being removed, so that is to consider.


lemire:
@geofflangdale : want to have a reviewing this?

lemire:
Anyhow, this looks legitimate so I am going to merge and work through any issue that might arise.

Thanks!

atomheartother:
@lemire neat, let me know if it actually breaks anything.

lemire:
It broke stuff but I fixed it. We now have CI so future PR will be better assessed.

PR #60, opened 2019-02-23
Build with both clang-6.0 and gcc-7.

### LINKED ISSUE #56: add cmake testing to our circleci setup
We are currently not testing using cmake with circleci...
https://github.com/lemire/simdjson/blob/master/.circleci/config.yml

TkTech:
Closes #56.

Two isolated workflow builds, one running clang-6 and the other gcc-7 using the Uubuntu 18.04 LTS as a base image. Runs with both the traditional make as well as CMake.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,82 @@
 version: 2
 jobs:
-  build:
-    docker: 
-      - image: gcc:8
+  "gcc":
+    docker:
+      - image: ubuntu:18.04
+        environment:
+          CXX: g++-7
     steps:
-      - checkout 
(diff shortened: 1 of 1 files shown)


lemire:
appveyor fails, but I think it is a time out issue.

TkTech:
I have done several dozen appveyor builds as part of the pysimdjson tests, haven't run into it just stopping in the middle of tests. Appveyor doesn't time out until 60 minutes from job start.

Almost looks like a kernel halt, instance just completely stopped.

lemire:
If we can't get a clean build, we will have to investigate.

PR #61, opened 2019-02-23
Fix minor problems

NewProggie:
Following is a first batch of smaller fixes concerning old style C-casts, const correctness, implicit casts and using delete functions wherever appropriate


lemire:
The PR appears to fail on continuous integration tests. There are problem building the code. Can you check those out?

See in particular https://cloud.drone.io/lemire/simdjson/26/1/2

lemire:
Looks fine. Let us wait for CI to run.

NewProggie:
> Looks fine. Let us wait for CI to run.

Yes. I've noticed this issues already yesterday, but didn't had the time to get to it right away.

lemire:
This now looks fine. This PR basically turns C-isms into C++-isms.

I'd feel better if someone else looked at this long list of (sensible) changes before I merged.  

Maybe @geofflangdale? Someone else?

geofflangdale:
I've looked at this change list. It generally looks very good. The only thing I find unpalatable, in the non-pointer case, is turning conditions like "if (foo(x))" into "if (foo(x) != 0u)" - is there a reason that this rather verbose form is being used?

lemire:
I spotted this as well. @NewProggie do you have a particular reason for the verbose "if (foo(x) != 0u)"? I don't think it is particular idiomatic, is it?

NewProggie:
> I spotted this as well. @NewProggie do you have a particular reason for the verbose "if (foo(x) != 0u)"? I don't think it is particular idiomatic, is it?

Yes, that seems reasonable. I've updated this PR and re-added the implicit bool conversions

PR #65, opened 2019-02-24
Use nothrow

WojciechMula:
None of allocation checks in `ParsedJson::allocateCapacity` would be executed, because `new` by default throws.

BTW I'm thinking about purpose of this method, and for me it looks rather like a constructor. I checked all use cases in your code and the pattern is the same: default-construct ParsedJson, then execute on that instance allocateCapacity. If all allocations were done in constructor you might use `std::unique_ptr` for memory management and also get rid of explicit `dealloce` method from API.


lemire:
Great. Merging. 

Regarding the allocate-as-constructor, I think this would be better handled as a separate issue and/or pull-request.

PR #67, opened 2019-02-25
Update CMakeLists.txt, .travis.yml and .drone.yml

xorz57:
Adds support for CPack so that you can make .deb and .rpm packages.
Also tried to improve `.travis.yml` and `.drone.yml`


lemire:
Looks good. Not that we have disabled travis since they occasionally use older CPUs and we have no obvious way to state requirements on the CPUs.

PR #76, opened 2019-02-26
CMake: Add version and soversion to library target properties.

cdluminate:
Closes: https://github.com/lemire/simdjson/issues/52

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,8 @@ endif()
 
 project(simdjson)
 set(SIMDJSON_LIB_NAME simdjson)
+set(SIMDJSON_LIB_VERSION "0.0.1" CACHE STRING "simdjson library version")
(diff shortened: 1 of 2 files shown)


lemire:
This is ridiculous: appveyor keeps on timing out. Restarting for luck?

lemire:
Looks great to me. I'll patch it up later to match with the release script.

lemire:
@cdluminate Honestly, I don't understand the soversion thing. We already have a library version. Why do we need more?

lemire:
The CMake says "If only one of both is specified the missing is assumed to have the same version number."

So why do we need a soversion in the first place?

lemire:
> So why do we need a soversion in the first place?

Anyhow, the practice we have is to bump the number by one with every minor release.

I think it would be far nicer if we could just tag the libraries with our version numbers, but this is apparently not practical.

PR #78, opened 2019-02-26
 Fix CRLF line terminators

mr700:
This converts the line terminators from CRLF to LF in two files (one auto-generated).

### DIFF
--- a/amalgamation.sh
+++ b/amalgamation.sh
@@ -119,16 +119,16 @@ echo "Giving final instructions:"
 CPPBIN=${DEMOCPP%%.*}
 
 echo "Try :"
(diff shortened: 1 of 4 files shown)


lemire:
Can you sync with master once more, and only include changes in 
include/simdjson/common_defs.h. The other one is automatically generated and not meant to be edited so I don't think we should worry about line endings.

lemire:
(Someone else must have modified include/simdjson/common_defs.h between the moment you did your work and me getting around to checking the PR.)

mr700:
There is no longer need to do it, it is already fixed in your branch...

lemire:
Sorry about that. I think that all I did was to add  https://github.com/lemire/simdjson/blob/master/.gitattributes

This seems to have magically resolved the issue.

PR #81, opened 2019-02-26
WIP: Change parse_json return type from bool to int

atomheartother:
This branch is my attempt to make `parse_json` behave as is standard for C functions, i.e. have it return 0 on success and an error code otherwise. **This is a work in progress, do not merge**.

# Major Changes
## simdjerr namespace
There's now a new header, located at `simdjson/simdjerr.h`. This header defines the `simdjerr` namespace, which contains an enum for return values. This enum is **weakly typed** (C-style), because making users `static_cast` to int seems like too much of a pain for the potential benefits.

Ideally we should gradually transition public-facing function returns to this enum, so that the user can get a meaningful message out of it later, rather than us printing to console.

## Removed parse_json logs on error, returned int instead 
`parse_json`, along with the functions it calls and checks the return value off, now don't print their error messages to `stderr`, they return a value scoped within `simdjerr` on error, and 0 on success.

**This breaks backwards-compatibility** as the old return value on success was 1. The README example has been updated accordingly.

# Small changes
- `json2json`, `parse` and `jsoncheck` have been  updated to match the new return values.
- Move two static inline functions for `unified_machine` and `find_structural_bits` from header files to their respective `.cpp` files. I don't see any reason to have these functions as inline static, they will cause warnings if they're included & unused and will inflate binary size unnecessarily.

# To do
- Triple-check that every tool and benchmarking tool shipped with this repo conforms to the new return values, I might not know them all.
- Add a static function to the `simdjerr` namespace which returns a string when given an error code, in order to allow for meaningful debugging for the user.
- Transition more functions to the `simdjerr` namespace?


atomheartother:
... what the hell, it looks like #47 was un-merged from `master` and those changes got stashed with my current changes. Ack.

atomheartother:
Closing this, I'll re-do it later because right now it looks like #47 was merged extremely strangely, some of my commits from back then are being re-added, even though i can clearly see they're already on `lemire:master`? I have no clue.

TkTech:
Can I make a few suggestions?

```
namespace simdjerr {
  enum errorValues {
    SUCCESS = 0,
    CAPACITY, // This ParsedJson can't support a document that big
    MEMALLOC, // Error allocating memory, most likely out of memory
    TAPE_ERROR, // Something went wrong while writing to the tape
  };
}
```

Is there a reason these are in their own namespace? They should probably be in the same namespace as simdjson (...which also needs to be namespaced).

Define error strings for each error code rather than a comment, so the rest of the code and people using simdjson as a library can get the error strings.

```c

const char* error_messages[] = {

}
```

PR #84, opened 2019-02-27
Replace _mm256_storeu_si256 with memcpy in stringparsing.h

EgorBo:
A minor change - `v` is basically `src` here and we can use memcpy instead which is faster (on my env at least). Benchmarks show 1% performance improvement 😄

### DIFF
--- a/include/simdjson/stringparsing.h
+++ b/include/simdjson/stringparsing.h
@@ -105,7 +105,8 @@ really_inline  bool parse_string(const uint8_t *buf, UNUSED size_t len,
     uint32_t bs_dist = trailingzeroes(bs_bits);
     // store to dest unconditionally - we can overwrite the bits we don't like
     // later
-    _mm256_storeu_si256(reinterpret_cast<__m256i *>(dst), v);
+    memcpy(dst, src, sizeof(__m256i));
+
     if (quote_dist < bs_dist) {
       // we encountered quotes first. Move dst to point to quotes and exit
       dst[quote_dist] = 0; // null terminate and get out


lemire:
@EgorBo Why would you expect it to be faster? The intrinsic should map to a single instruction... or am I wrong?

I would expect the memcpy to be super efficient as well, but I am not sure how it could be *more* efficient.

Thoughts?

cc @geofflangdale

geofflangdale:
I don't understand why this would be faster, and suspect a 'phase of the moon' speedup. Possibly the compiler knows more about memcpy's semantics than the intrinsic? The thing that worries me about the change is that we're guaranteed that the intrinsic maps to an instruction, while we are taking "who knows what" when we use memcpy. It makes me nervous, even when the work on fixed-size memcpys has been generally execellent.

I would prefer that we understand why this is faster at a code level than make a bet on lots of unknown compilers and runtimes.

EgorBo:
Yeah I agree, I do not fully understand why, the only thing I have is this godbolt https://godbolt.org/z/ughe5U so yes it doesn't make sense.

geofflangdale:
https://godbolt.org/z/qkw7Or with gcc is... weird. Both sides generate weird code (memcpy  insists on SSE, while the "intrinsic" version is using AVX2 registers but insisting on working 128 bits at a time.

TkTech:
When you benchmarked it locally did you use /O3 or /O2?

EgorBo:
@geofflangdale well, maybe gcc decided that sse is faster than avx + vzeroupper :) 
@TkTech /O2 (MSVC)

TkTech:
I could reproduce this, but only on the short tests in jsonexamples.  i7-4470k, msvc 2017 /AVX2 /2, forced affinity to core 0.

On a larger test (sandiego.json, 181MB, lots of strings), `_mm256_storeu_si256` variant was always faster, but not by much.

P-State magic?

PR #224, opened 2019-07-25
[draft] Operate on 256-bit bitmasks instead of 64

jkeiser:
Right now simdjson stage 1 uses 256-bit wide SIMD to process bytes, but then uses 64-bit operations to process the resulting masks. This PR walks over the input in 256-byte chunks, generating 256-bit-wide SIMD masks and using 256-bit-wide SIMD to process said masks. Right now, only odd backslash detection is converted.

**This is a draft.** I'm generating this PR in the hopes someone can glance at it and see something I haven't. I'm benching it by parsing [one of the larger input files](https://github.com/lemire/simdjson/blob/master/jsonexamples/marine_ik.json) in a loop 1000 times on my Windows machine, and it's not slower, but it's not faster either. I'm puzzled as to why--it definitely seems like fewer instructions should be running. Hopefully I'm doing something silly (I'm new at both SIMD and low-level performance), but even if not, perhaps this will suggest other directions to go.

## The main code changes

* [find_structural_bits](https://github.com/lemire/simdjson/pull/224/files#diff-df04b44cb07019991ed2e6c926208a5aR1222-R1241) now loops over 256-byte chunks. It then [processes the input in 64-byte chunks](https://github.com/lemire/simdjson/pull/224/files#diff-df04b44cb07019991ed2e6c926208a5aR1160-R1163) (which actually internally ends up operating on 32-byte chunks, because that's 256 bits).
* All byte input is processed and turned into bitmasks [in one place](https://github.com/lemire/simdjson/pull/224/files#diff-df04b44cb07019991ed2e6c926208a5aR1120-R1139). Particularly, I moved find_whitespace_and_structurals up to the top so that we could process each 32 byte input block once. This did not make any performance difference on my AVX-2 machine, however, so it can be moved back just as easily.
* [`find_odd_backslash_sequences_256`](https://github.com/lemire/simdjson/pull/224/files#diff-df04b44cb07019991ed2e6c926208a5aR715-R735) has most of the 512-bit logic in it. I haven't converted most of the other functions, yet, but this has the highest concentration of mask logic, so we should be seeing benefits here. It has nearly the same code as before, just on simd_bitmasks rather than uint64_t. It does change how overflow is handled, a bit, but keeps that the same conceptually as well.
* There is no ARM implementation, as I don't presently have a spot to test it; I'll look at it once this is actually running faster than the 64-bit version.
* Unrolling loops and avoiding lambdas did seem to be important for perf, fyi--I'm assuming msbuild just kinda sucks at those optimizations.

## simd_bitmask

[`simd_bitmask<instruction_set T>`](https://github.com/lemire/simdjson/pull/224/files#diff-df04b44cb07019991ed2e6c926208a5aR55-R59) defines a 256-bit mask: `__m256` on AVX-2 and 2 `__m128i`'s on SSE4.2.

* It has [associated functions](https://github.com/lemire/simdjson/pull/224/files#diff-df04b44cb07019991ed2e6c926208a5aR412-R585) for 512-bit &, |, ^, ~, << and +. For example:
  ```c++
  really_inline simd_bitmask<instruction_set::avx2> operator&(const simd_bitmask<instruction_set::avx2> a, const simd_bitmask<instruction_set::avx2> b) {
      simd_bitmask<instruction_set::avx2> result;
      result.mask = _mm256_and_si256(a.mask, b.mask);
      return result;
  }
  ```
* << and + (which can overflow) have an overflow version that takes `uint64_t &overflow` parameter, uses it on the way in, and sets it on the way out. For example, [`shl_overflow`](https://github.com/lemire/simdjson/pull/224/files#diff-df04b44cb07019991ed2e6c926208a5aR485-R501) pulls the overflow in as the lowest bit, and then sets the overflow based on the highest bit.
* `add_overflow` presently just runs a 64-bit add against each of the 4 64-bit values, tossing the overflow from one to the next; I've thought of algorithms to do it in SIMD, but until we hit 512 it feels like the number of instructions is lower in 64 bits. (It's possible, however, that data dependency between each of the 4 adds and their overflows will cause the processor to stall a bit; if that's the case, a more parallel operation might be a win?


jkeiser:
It looks like one of the adversarial tests failed (not surprising, given my changes necessarily have to touch the overflow logic). I had assumed `basictests` and `jsoncheck` were the whole set of tests (they pass when I run them), so I'll look into this later. Meantime, editing the description here with a bit of a guide so people can see what I've modified and maybe venture guesses as to why we're not seeing an actual improvement.

jkeiser:
Added a guide to the changes in the PR description.

lemire:
@jkeiser Fantastic.

There is an upcoming change to account for runtime dispatch, but it will be interesting to investigate your proposal.

@geofflangdale and I discussed something of the sort, though I don't think it was ever implemented.

Let us assess this...!

geofflangdale:
Good work. I laid out the ideas for this transformation in a document somewhere but could not persuade myself to try it.

There are a couple of straightforward ways to do parallel prefix XOR with SIMD also. You can use a couple PSHUFBs to look up "high nibble" and "low nibble" tables of XOR values, which then allows you to have bytewise parallel prefix XOR. You then take bit 7 out of every byte, do parallel prefix XOR on *that* (probably using the current method), then transform that back from a bit-mask (32 or 64 bits to 256 or 512 bytes), and XOR that value over the top of the bytes. Now you have parallel prefix XOR.

This will be way easier on AVX512-GFNI.

Generally the wide add operation will also be easier on AVX-512 - you can do the compares to check for overflow and work in the kregs. This could be done in AVX2, but I agree that it's unclear whether the 256-bit version is worth it, though. Generally the number of instructions and the length of the dependency chain can both be important in the codebase, so it might be worth experimenting.

lemire:
Note that we are almost done with the move to runtime dispatch that changes the code, so there will be some non-trivial merge needed later to update this PR.

jkeiser:
Yeah, I'd be happy to convert it over once that's done. At any rate, this patch isn't faster than the 64-bit-wide version yet, so I've got me a date with a profiler :)

lemire:
@jkeiser 

I am hoping you will like the new code that @ioioioio will push for runtime dispatch.

lemire:
Version 0.2.0 of simdjson was just released.

lemire:
@aqrit Clever!

jkeiser:
Sweet, I'll check it out today!

jkeiser:
Wow, I didn't even realize you could do email replies. Neat.

jkeiser:
I'm going to close this one in favor of [the corresponding issue](https://github.com/lemire/simdjson/issues/261), so that I don't sully the clean PR list while working on this. I'll link here, and @aqrit I haven't lost your change :)

PR #262, opened 2019-08-06
Use direct call to templated flatten_bits instead of if

jkeiser:
Templatizing flatten_bits like the rest of the code the right thing to do regardless whether the if is causing a perf problem, methinks.

NOTE: I haven't perf tested or Linux/ARM-tested this, yet. Relying on CI for the latter.


lemire:
We may need to update the amalgamation script.

cc @ioioioio

jkeiser:
Ah. I was expecting it to read `stage1_find_marks.cpp` to find out how to include the .h files; I bet it doesn't do that. I'll see if I can fix it up to do that tonight ...

lemire:
@jkeiser The amalgamation script is *not* smart.

lemire:
@jkeiser  I can fix the amalgamation script for you if you want. Let me know.

jkeiser:
I'll take a look when I get home and have a Linux :)

jkeiser:
OK, moving flatten_haswell.h to CFILES fixes the problem. Gives me the willies though :(

ioioioio:
I've just noticed you only have to add really_inline to find_structural_bits_64 on that branch and it fixes the performance regression.

lemire:
@jkeiser  Your PR appears to fail under Visual Studio...(see appveyor)

jkeiser:
wow, really_inline fixes it?

Also, of *course* my fix that I only tested on Linux fails on my host OS :) Will fix shortly.

jkeiser:
Can't believe I *lost* really_inline. argh.

jkeiser:
Regardless of reason, it is criminal user experience that any compiler, especially one that has existed for decades, is willing to produce this error:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2664	 'uint64_t simdjson::cmp_mask_against_input<simdjson::Architecture::HASWELL>(simdjson::simd_input<simdjson::Architecture::HASWELL>,uint8_t)': cannot convert argument 1 from 'simdjson::simd_input<simdjson::Architecture::HASWELL>' to 'simdjson::simd_input<simdjson::Architecture::HASWELL>'	simdjson	C:\Users\john\Source\simdjson\include\simdjson\stage1_find_marks_common.h	27	
```

jkeiser:
OK, this latest push compiled (using CI to double check). I haven't checked perf myself, but I did add `really_inline` back to find_structural_bits_64.

I am not happy with the .h file being in `ALLCFILES`, but I think the right solution might be a reorganization of headers so that interfaces are in `include/simdjson` and implementations are in `src`. I say "I think" because I haven't got a good grasp on where we intend our inline boundaries to be: my intuition says that we're probably OK with the outer parse function being non-inline and all the inline stuff being implementation.

Fixing that can and should be done separately and after; the fix isn't so toxic that it's worth keeping the perf regression around.

ioioioio:
I tested the performances. Everything seems fine.

lemire:
I agree that we should separate out the header files. This was discussed several times. Never implemented. I will open annissue.

jkeiser:
@lemire FYI tests are passing and it significantly improves performance over master on my machine as well.

lemire:
I read it over and it is nice and simple. Merging.

PR #270, opened 2019-08-11
check for performance degradation in CI

jkeiser:
This adds a performance degradation test in .drone.yml (hopefully, we'll see what happens when the PR runs :)). It checks out `master` and compares to this branch, and fails if it degrades more than 10%. Advice requested on what tolerance to use here--0% doesn't work because variance. I expect we'll also want to track it over time to make sure we don't consistently lose 1% here, 2% there until it adds up

To try to mitigate variance in performance numbers (especially in CI), I added a few things. Suggestions / feedback welcome:

* The `checkperf` is a separate step, with the hopes that we can re-run it if we believe we got a spurious result.
* We compile and measure master and current-branch performance in the same machine / container / CI run, eliminating the possibility of being assigned different hardware and strengthening the chance that the VM / container is shared with the *same* neighbor VMs / containers in the cloud (with similar workloads, one hopes).
* We build `parse.exe` for both master and current and *then* run the resulting .exes one after the other, minimizing the time between runs. Hoping that "noisy neighbors" will often be consistent enough for short periods of time that either *both* runs will be interfered with the same, or *neither* will.
* `parse.exe` runs a few "warmup" iterations where it throws away the result, to ensure the instruction caches are loaded. Not really sure what kind of difference this will make, but I've seen this done in other benchmarks, and it didn't seem wrong here :)
* `perfdiff.exe` runs the performance difference a couple of times, interleaved, and averages the results. Mainly to increase the chances that a super noisy neighbor affects both runs. Not really sure how much it impacts.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -12,6 +12,10 @@ steps:
   - make
   - make quiettest
   - make amalgamate
+- name: checkperf
+  image: gcc:8
+  commands:
+  - make checkperf
 ---
 kind: pipeline
 name: arm64
@@ -27,6 +31,10 @@ steps:
   - make
   - make quiettest
   - make amalgamate
+- name: checkperf
+  image: gcc:8
+  commands:
+  - make checkperf
 ---
 kind: pipeline
 name: stylecheck
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,9 @@ $(JSON_INCLUDE) $(SAJSON_INCLUDE) $(RAPIDJSON_INCLUDE) $(JSON11_INCLUDE) $(FASTJ
 parse: benchmark/parse.cpp $(HEADERS) $(LIBFILES)
 	$(CXX) $(CXXFLAGS) -o parse $(LIBFILES) benchmark/parse.cpp $(LIBFLAGS)
 
+perfdiff: benchmark/perfdiff.cpp
+	$(CXX) $(CXXFLAGS) -o perfdiff benchmark/perfdiff.cpp $(LIBFLAGS)
+
 statisticalmodel: benchmark/statisticalmodel.cpp $(HEADERS) $(LIBFILES)
 	$(CXX) $(CXXFLAGS) -o statisticalmodel $(LIBFILES) benchmark/statisticalmodel.cpp $(LIBFLAGS)
 
--- a/Makefile
+++ b/Makefile
@@ -126,6 +126,9 @@ parse: benchmark/parse.cpp $(HEADERS) $(LIBFILES)
 perfdiff: benchmark/perfdiff.cpp
 	$(CXX) $(CXXFLAGS) -o perfdiff benchmark/perfdiff.cpp $(LIBFLAGS)
 
+checkperf:
+	bash ./scripts/checkperf.sh
+
 statisticalmodel: benchmark/statisticalmodel.cpp $(HEADERS) $(LIBFILES)
 	$(CXX) $(CXXFLAGS) -o statisticalmodel $(LIBFILES) benchmark/statisticalmodel.cpp $(LIBFLAGS)
 
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -6,3 +6,4 @@ target_include_directories(${SIMDJSON_LIB_NAME}
 
 add_cpp_benchmark(parse)
 add_cpp_benchmark(statisticalmodel)
+add_executable(perfdiff perfdiff.cpp)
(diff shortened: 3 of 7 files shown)


lemire:
Great work. 

I think we should merge this... Before we do, here are my comments which you may consider...

1. You compare against master... that's a moving target. As you say, we can progressively get slower without noticing.  Don't we also want to have a fixed target? Currently our last release v0.2.1. It is possible [to automatically find the last release](https://stackoverflow.com/questions/17414104/git-checkout-latest-tag) though I guess that aiming to be never much slower than version 0.2.1 is already pretty good. If we use a fixed reference release, then this may need to be manually updated.


2. I agree that it is really hard to set a tight absolute threshold on throughput. Sadly, under something like drone, we can't easily use other, more reliable, measures like "instruction counts". However, it might be easy to add statistical tests that are more robust.  For example, suppose that the throughput distribution is the same (as a probability distribution), then half the time, the new code should be faster. Now, if you have 6 interleaved attempts, what is the probability that in all 6 instances, the new code will be slower? It should be 1/2^6 or less than 0.5%. That should only rarely happen because we don't issue 100s of PRs... and when the test fails, we can ask for a re-run in any case. (And ultimately, we may merge even if there is a slightly performance regression, however, the author will now be aware of the regression.) That is, if you do 6 comparisons and all 6 of them are in favor of the old code, maybe this warrants further investigation.

jkeiser:
Yeah, I think I'm fine with having a reference release so we always know.
Regarding the "try it 6 times," is the idea that we would essentially set
the bar at 0%, then retry the perf test on failure up to 6 times and only
mark it failed if it failed all 6?

I'm on a boat at the moment so can't update it just now, but I'm on board.

Note: I am seeing something like 5-10% variance on my machine, though I'm
running other stuff and I'm also running this in a docker container from my
host windows os since it's Linux focused. One of these days I'll get myself
dual booted :)

--John

jkeiser:
I'm also good with finding the latest release, though I'm finding bash
scripting somewhat painful :) I'd rather have written perfdiff.exe in a
scripting language (Ruby Perl python or anything) or in bash, too, but we'd
need to get said language into our docker images before we could go there.

jkeiser:
We also need to run this for sse4.2 vs avx2.

lemire:
> We also need to run this for sse4.2 vs avx2.

Granted, though have something today, even if it is only partial, is going to be helpful.

lemire:
> we'd need to get said language into our docker images before we could go there.


See this bit we already use...

```
kind: pipeline
name: stylecheck

platform:
  os: linux
  arch: amd64

steps:
- name: Build and Test
  image: ubuntu:18.04
  commands:
    - apt-get update -y
    - apt-get install -y python clang-format
    - ./style/run-clang-format.py -r include/ benchmark/ src/ tests/
```

We have an automated style check in Python.

We allow failure (so it is meant for human consumption).

lemire:
>Yeah, I think I'm fine with having a reference release so we always know.
Regarding the "try it 6 times," is the idea that we would essentially set
the bar at 0%, then retry the perf test on failure up to 6 times and only
mark it failed if it failed all 6?

Yes! Basically, this avoids the need to have a made up "10%".

> I'm on a boat at the moment so can't update it just now, but I'm on board.

Hopefully you are not on a boat looking at your smartphone? My wife says that I have to stop acting that way. It is impolite. She says.

> Note: I am seeing something like 5-10% variance on my machine, though I'm
running other stuff and I'm also running this in a docker container from my
host windows os since it's Linux focused. One of these days I'll get myself
dual booted :)

I thought Microsoft did a decent job at the Linux subsystem? I found it too hard to install (it should just be "click install and go"), but it sure works well when it works.

I'd gladly do more to Windows CI...

jkeiser:
To be clear, I'm also OK with merging this and fixing the other bits. Never let the perfect be the enemy of the good. I'll file an issue with these comments if you merge before I change things (which probably isn't tonight).

While I do have permission to merge now, I don't feel I'm *quite* integrated enough with the project to be actually doing it yet, nor would I generally feel comfortable merging my own PRs most of the time anyway :)

jkeiser:
Re boats, my wife agrees with yours! I had lots of fun on the boat, but sometimes reality gets a little repetitive and I sink back into the realm of the mind :)

jkeiser:
This fixes #260.

lemire:
Ok. I will merge and maybe tweak a bit.

lemire:
Ok: I think I am not going to touch it for now given the next issue.

PR #284, opened 2019-08-16
Lower the bar for performance check

jkeiser:
Since we run the performance check 8 different times in different CIs, we have to run a 7th interleaved attempt to keep the probability of false failure in the same range. I also increased the threshold to 0.5% while we were at it. False failure should be few and far between with this change, but should still reliably detect reduced performance above a certain noise level.


jkeiser:
Feeling good about all this for the moment, I'll start back up on some actual improvements next :)

PR #324, opened 2019-10-06
case insensitive move_to_key

jplauri:
As per [issue 323](https://github.com/lemire/simdjson/issues/323), this PR introduces a case insensitive key lookup for the iterator. The proposal refactors `move_to_key` into a private `move_to_key_impl` which takes not only the key but a comparator. This way, there's no code duplication and both `move_to_key` and the newly added `move_to_key_insensitive` are added cleanly.

In principle, there should be zero overhead from the comparator.

### DIFF
--- a/include/simdjson/parsedjsoniterator.h
+++ b/include/simdjson/parsedjsoniterator.h
@@ -339,15 +339,15 @@ void ParsedJson::BasicIterator<max_depth>::move_to_value() {
 
 template <size_t max_depth>
 bool ParsedJson::BasicIterator<max_depth>::move_to_key(const char *key) {
-  return move_to_key_impl(
+  return this->move_to_key_impl(
       key, [](const char *x, const char *y) { return strcmp(x, y) == 0; });
 }
 
 template <size_t max_depth>
 bool ParsedJson::BasicIterator<max_depth>::move_to_key_insensitive(
     const char *key) {
-  return move_to_key_impl(
-      key, [](const char *x, const char *y) { return strcmpi(x, y) == 0; });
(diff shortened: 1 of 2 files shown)


jplauri:
Oh, I suppose that `strcmpi` is Windows-specific while UNIX systems have [`strcasecmp`](https://en.wikibooks.org/wiki/C_Programming/C_Reference/nonstandard/strcasecmp). Is there anything in the code currently that is differentiating between the two, like picking a function to be called?

jkeiser:
I can't quite tell if the two are identical for > and <, but it's pretty likely they are identical for ==. You could define a strcmpeq in [portability.h](https://github.com/lemire/simdjson/blob/master/include/simdjson/portability.h) that returns `bool` if they are different, or just use `strcasecmp` and something like this:

```c++
#ifdef _MSC_VER
#define strcasecmp stricmp
#endif
```

jplauri:
Great, thanks John! I'll update this in a few hours.

lemire:
Ok. I am going to merge to a local branch for performance assessment and then merge into master if I don't see any problem.

lemire:
I accidentally merged into master. I think that there is a performance regression. I'm investigating.

jplauri:
> I accidentally merged into master. I think that there is a performance regression. I'm investigating.

Interesting. So do you mean that the new `move_to_key` is slower than the old one?

lemire:
Yes. This is fixed in master.

lemire:
Yes, there was a measurable performance regression.

jplauri:
OK, good to know. Thanks for checking & thanks for the work!

PR #375, opened 2019-11-21
Backport to ARMv7

jkeiser:
@lemire this is your ARMv7 branch that adds CI to @easyaspi314 's PR #330. I'm just opening up a PR so we can see what CI says about it and maybe merge.


jkeiser:
Dumping this PR in favor of the original, since it didn't work.

PR #376, opened 2019-11-22
Make threads optional in the cmake build

pauldreik:
After recent discussions on threads in the cmake build, this cleans up the cmake
threading handling and makes it configurable.

Update: the CI jobs seem to work, this PR is ready for review!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -150,6 +150,8 @@ workflows:
       - clang-sse-dynamic
       - clang-sse-static
       - clang-sse-sanitize
+      - gcc-avx-threaded
+      - gcc-avx-unthreaded
 
       # TODO add windows: https://circleci.com/docs/2.0/configuration-reference/#windows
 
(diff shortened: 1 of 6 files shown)


piotte13:
+1

pauldreik:
This is ready for review.

.circleci/config.yml
```diff
@@ -139,6 +150,8 @@ workflows:
       - clang-sse-dynamic
       - clang-sse-static
       - clang-sse-sanitize
+      - gcc-avx-threaded
+      - gcc-avx-unthreaded
```
piotte13: Do you suggest that doing the same on clang is not necessary?

pauldreik: No, testing all combinations is good except for the increased test time. I added the clang variants!

piotte13: Thanks Paul.  Good work.

PR #433, opened 2020-01-05
Using  standard C++

DBJDBJ:
Let the debate begin :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+visual_studio
+
 /benchbranch/
 /submodules/
 
--- a/tests/jsonstream_test.cpp
+++ b/tests/jsonstream_test.cpp
@@ -2,6 +2,13 @@
 #include "simdjson/jsonparser.h"
 #include <simdjson/jsonstream.h>
 
+/*
+Originally in the C standard library as <limits.h>.
+Part of the type support library, in particular it's part of the
+C numeric limits interface.
+*/
+#include <climits>
+
 // this is MSVC STD LIB code
 // it actually does not depend on C++20 __cplusplus 
 // which is yet undefined as of 2020 Q1
--- a/tests/jsonstream_test.cpp
+++ b/tests/jsonstream_test.cpp
@@ -1,154 +1,263 @@
-#include <cassert>
-#include <cstring>
-#ifndef _MSC_VER
-#include <dirent.h>
-#include <unistd.h>
-#else
-// Microsoft can't be bothered to provide standard utils.
-#include <dirent_portable.h>
-#endif
-#include <cinttypes>
-
-#include <cstdio>
-#include <cstdlib>
-#include <simdjson/jsonstream.h>
 
 #include "simdjson/jsonparser.h"
+#include <simdjson/jsonstream.h>
 
-/**
- * Does the file filename ends with the given extension.
- */
-static bool has_extension(const char *filename, const char *extension) {
-    const char *ext = strrchr(filename, '.');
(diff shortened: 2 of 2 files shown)


DBJDBJ:
image


PR build fails because it uses C++17 feature [not present in G++ 7.x](https://stackoverflow.com/questions/45867379/why-does-gcc-not-seem-to-have-the-filesystem-standard-library)

@lemire Can you please define minimal versions of the clang, msvc and gnuc .

lemire:
I think that all main compilers support fully C++17 at the language level. For the library features you require, we need GNU GCC 9:

https://gcc.gnu.org/gcc-9/changes.html

I'm still on GNU GCC 8 and the CI on on GNU GCC 7 or older (the project started with GNU GCC 7).

It feels a bit drastic to require GNU GCC 9 when it was just released (2019). I don't know any major Linux distribution defaulting on it yet.

Could you tweak your PR so that instead of replacing our existing code, you only make obsolete our Visual Studio hack (dirent.h)?

We will migrate to the new filesystem API for sure, but it sounds like we should give ourselves time. So we want some hybrid solution meanwhile.

DBJDBJ:
@lemire  -- I do require C++17 :)   In this instance, it makes OS agnostic "hacks" redundant.  Thus making code more resilient.

Can we agree on 8.x for both gnuc and clang?   For the development and for CI. 

[Here is a good overview](https://en.cppreference.com/w/cpp/compiler_support) on what is supported where.

Mind you, I am not a big fan of `std::` lib. For tight space projects, I am using [this very clean C implementation](https://github.com/mattiasgustavsson/newpixie/blob/master/source/pixie/dir.h) of dir all in a single simple header, for both POSIX and WIN32. Will using it elsewhere, provoke a lot of changes in simdbjson?

In this PR, I have replaced the existing code to show a few important points:

-- C++17 (and std:: lib) makes for shorter and more resilient code
-- simdbjson core functions can and [should be exceptions free](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1640r1.html). 
-- std::string usage can and should be optimized (read: drop it if you can)

Idea is to isolate all of that in one cpp file. This is testing code. Not core.

lemire:
I have moved us from gcc:7 to gcc:8, please see this PR:

https://github.com/lemire/simdjson/commit/6706d6053e639bc362f4ab73508e243305fca04f

I recommend you sync your fork with master to take this change into account.

It still won't help. GNU GCC 8 does not come with filesystem support integrated, a stable filesystem support is a feature of GNU GCC 9.  You can "hack" filesystem with GNU GCC 8, but I think we have more urgent things to do than to play with experimental features.

lemire:
> I have replaced the existing code to show a few important points

I agree with the points you are making and find your PR valuable. 

I am arguing back in terms of practicality. I think it is too harsh to demand that the developers of this library move to the GCC 9 (which is not even a year old). It needs to work with GCC 8.

DBJDBJ:
I re-forked yesterday? for example `.drone.yml` says `image: gcc:8` ... 

Ok, I assume I can "request" gnuc 8.x in my code in this PR.

I repeat :) I am not very keen on std:: lib.  that includes `<filesystem>` which I think is abomination because of the dual personality of using exceptions and a non-exceptions version for each function in there.

I can use: https://github.com/mattiasgustavsson/newpixie/blob/master/source/pixie/dir.h , at least in the `~/tests/` folder

DBJDBJ:
closing due to re-fork due to move to gnuc 8.x, clang 8.x

PR #434, opened 2020-01-07
dirent portable refresh 2019

DBJDBJ:


lemire:
Please revert the non-relevant submodule changes?... otherwise it is too hard to figure out what is happening.

DBJDBJ:
@lemire the whole solution is just too big and confusing non-standard VisualStudio solution with 20 cmake projects ...  Now I am forced into GIT CLI sorcery :)   

Not much is "happening" in this PR besides adding the latest, much cleaner `dirent.h` from Tony's repo ... All are in the "windows" folder.

After this PR, I have made another branch. If I `git revert bc951e8` will that new branch stay intact?

DBJDBJ:
Ok, somehow (it seems) I did it?

DBJDBJ:
So this is the situation in my VS2019

image

(Obviously project names mean not a lot, to use mere co-workers)

And this is the situation on one of the numerous CI platforms where PR fails.

image

What is confusing CLANG 6 is used? Not 8. Also, the error message is a "mystery" since a few other cpp's in the same folder have the same include which works? 

And in development mistery leads to misery ... as we all know.

Is it something I have done?

DBJDBJ:
Now the `dependancies/*` are failing .. I give up.

lemire:
You are changing too many things at once...
https://github.com/lemire/simdjson/pull/434/files

DBJDBJ:
@lemire there is some fundamental misunderstanding going on here. 

 I have changed two files and added one. 

I am not GIT CLI aficionado, neither I am a CMake ninja. Now. How do I get rid of those `dependancies/*` ? Obviously my "revert" attempt did not work ...

DBJDBJ:
will re-fork once  more

lemire:
> I am not GIT CLI aficionado

I recommend using git in the command line first. Using git in a GUI without first understanding it well enough to use it in a command line is a recipe for disasters.

DBJDBJ:
Now I am using Visual Code for GIT book-keeping --- with a hefty dose of CLI shenanigans.
VS2019 just for edit/compile/test

In any case VS sln (aka solution) is too big and project names are confusing. IMHO "workbench" , "testing", "fuzz" and "core" should be kept separately.  Even just renaming them would make people understand what is going on :

workbench_*
testing_*
core_*
fuzz_*

Dividing all in separate repos? Why not.  `lemire` should have github organization `simdbjson` with all of the above nicely laid out in separate repos.

> The ultimate goal is to be able to use simdjson (core) as a git submodule. 

For example. In the root folder of the project
```
git submodule add "https://github.com/lemire/simdjson/simdjson.git"
git submodule init
```
And then:
```cpp
// inside main.cpp
#include "simdjson/simdjson.h"
```
Stop everything and do this first :)

ps: simdjson + [hugo  ](https://georgecushen.com/create-your-website-with-hugo/ )= web site ?  Yes, that is on the horizon too ...

lemire:
1. Yes, we need a website.

2. I don't expect users to work from our repository. This repository is for core developers. The simdjson library is package in various ways (vpkg, debian, single-header).

3. Setting up an org is a good idea, but I don't think that's the next step. We have many, many important engineering tasks to complete for release 0.3, which is way behind schedule. Web sites, new org, renaming things... that's all secondary to core development.

PR #454, opened 2020-01-20
padded string memory leak

DBJDBJ:
Also -- get_corpus  standard implementation  +  moved to header

jsonioutil.cpp to be removed


lemire:
Can you merge with master and report performance results with the new get_corpus_benchmark? Ideally, we would like better performance.


https://github.com/lemire/simdjson/pull/456

DBJDBJ:
> Can you merge with master and report performance results with the new get_corpus_benchmark? Ideally, we would like better performance.
> 
> #456

1.  padded_string is actually streamlined (see the title?)  not `get_corpus()`. `get_corpus()` is just a side effect.
2. detrimental influence of exceptions on performance and size is [proven](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1640r1.html),  thus I conclude `simdjson` must not use exceptions. That also includes current `get_corpus()`.  this new 'get_corpus` uses much less exceptions **and** is very malleable to optimizations.
3. I have done as much as I can in my spare time.

lemire:
Ok. I ran the benchmarks...

Current speeds...(master)
```
$ ./get_corpus_benchmark_master jsonexamples/gsoc-2018.json 
file size: 0.00309928 GB
average speed: 4.25454 GB/s
min speed    : 2.79942 GB/s
max speed    : 5.86556 GB/s
```


Speed after merging this PR:

```
$ ./get_corpus_benchmark jsonexamples/gsoc-2018.json 
file size: 0.00309928 GB
average speed: 0.364734 GB/s
min speed    : 0.332593 GB/s
max speed    : 0.398785 GB/s
```

So, as you can see, it comes with more than a 10x performance penalty.

lemire:
I am pretty sure that the following is to be avoided at all cost if we care about performance...
```
   ifstream ifs_(filename.data());
   string str_(istreambuf_iterator<char>{ifs_}, {});
```

DBJDBJ:
image

@lemire

I expected that. My bench is [here](https://github.com/dbj-systems/simdjson_research) . 
I have added (in there) "no exceptions" `get_corpus()` version.  I do think the immediate roadmap of `simdjson` is very simple 3 steps:

1.  do not use try/throw/catch
2. do not use <iostream>
3. do not output to terminal/console 

> So, as you can see, it comes with more than a 10x performance penalty.

This always happens when one is using std lib. Good, we have this benchmark. 

I emailed you a few days (weeks) ago suggesting re-developing in C11, and if not then being economical on std lib usage. I repeat: I favor C++ core language, not std lib. 

The focus of my last PR  is padded_string, not `get_corpus()`, I will leave `get_corpus()` as it was (i.e. unchanged as it was before) of course.

DBJDBJ:
```cpp
  // move assignment
  padded_string &operator=(padded_string && other_ ) {
    using namespace std;
    std::swap(data_ptr, other_.data_ptr);
    std::swap(viable_size, other_.viable_size);
    // free the payload
    // remove this and you will have a mem leak
    simdjson::aligned_free_char(other_.data_ptr);
    other_.data_ptr = nullptr;
    other_.viable_size = 0;
    return *this;
  }
```
Usually closing the major mem leak makes "other things" pop-up elsewhere. On my WIN10 pro machine, all is fine beside two tests in the Release build. 
image

include/simdjson/padded_string.h
```diff
@@ -24,7 +35,9 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
     return nullptr;
   }
 #endif // NDEBUG
-  memset(padded_buffer + length, 0, totalpaddedlength - length);
+  
+    // ? -> memset(padded_buffer + length, 0, totalpaddedlength - length);
```
lemire: Why would you remove the memset?

DBJDBJ: if not strictly required that will improve the performance. What do you think? Although on -O3 that `memset` should be instant. Perhaps.  I am not sure. Testing as we speak.

DBJDBJ:
@lemire  this is all being done exactly like that, I have removed the "double free" etc .... give me a few more hours, I am testing it in isolation here ...  

In the mean time can you help me understand these two "failures"  ?

include/simdjson/jsonioutil.h
```diff
@@ -17,16 +17,18 @@ namespace simdjson {
...
-// caller is responsible to free (aligned_free((void*)result.data())))
 //
 // throws an exception if the file cannot be opened, use try/catch
+//
+//      padded_string p{} ;
 //      try {
 //        p = get_corpus(filename);
 //      } catch (const std::exception& e) {
-//        aligned_free((void*)p.data());
+//          p.~padded_string() ;
```
lemire: Calling the destructor in such a way almost ensures that the destructor will be called twice.

include/simdjson/padded_string.h
lemire: Have you determined that it is not required?

src/jsonioutil.cpp
```diff
@@ -1,36 +1,36 @@
 #include "simdjson/jsonioutil.h"
-#include <cstdlib>
-#include <cstring>
-#include <climits>
+ #include <cstdlib>
```
lemire: Please don't introduce unnecessary white-space changes. You have taken this one file and added a leading space to all but the first line. Please undo this change.

src/jsonstream.cpp
```diff
@@ -3,6 +3,7 @@
 #include "simdjson/isadetection.h"
 #include "jsoncharutils.h"
 
+ #include <stdexcept>
```
lemire: Please put the standard headers first.

include/simdjson/padded_string.h
```diff
@@ -33,77 +46,156 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
-  char *data() const  { return data_ptr; }
+  char *data() const { return data_ptr; }
 
 private:
+    // no copying
   padded_string &operator=(const padded_string &o) = delete;
   padded_string(const padded_string &o) = delete;
 
-  size_t viable_size;
+  size_t viable_size{};
```
lemire: This looks like an unnecessary change.

include/simdjson/padded_string.h
```diff
@@ -33,77 +46,156 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
   }
 
-  size_t size() const  { return viable_size; }
+  size_t size() const { return viable_size; }
 
-  size_t length() const  { return viable_size; }
+  size_t length() const { return viable_size; }
 
-  char *data() const  { return data_ptr; }
+  char *data() const { return data_ptr; }
```
lemire: This looks like a white-space only change, please avoid those.

include/simdjson/padded_string.h
```diff
@@ -33,77 +46,156 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
+      data_ptr = nullptr;
+      viable_size = size_t(0);
+    }
   }
 
-  size_t size() const  { return viable_size; }
+  size_t size() const { return viable_size; }
 
-  size_t length() const  { return viable_size; }
+  size_t length() const { return viable_size; }
```
lemire: This looks like a white-space only change, please avoid those.

include/simdjson/padded_string.h
```diff
@@ -33,77 +46,156 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
   ~padded_string() {
+    if (data_ptr != nullptr ) {
       aligned_free_char(data_ptr);
+      data_ptr = nullptr;
+      viable_size = size_t(0);
+    }
   }
 
-  size_t size() const  { return viable_size; }
+  size_t size() const { return viable_size; }
```
lemire: This looks like a white-space only change, please avoid those.

include/simdjson/padded_string.h
```diff
@@ -33,77 +46,156 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
+  padded_string b("B");
+       a.swap(b) ;
+         printf("%s", a.data() ); // "B"
+  */
+  void swap( padded_string &right_) {
+    // auto &left_ = *this;
+    std::swap(*this, right_);
   }
 
+  /*
```
lemire: Please don't include commented dead code in a PR to be reviewed.

include/simdjson/padded_string.h
```diff
@@ -33,77 +46,156 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
+    if (!ps_.reset(length_))
+      return ENOMEM;
+
+    std::rewind(fp);
+    size_t readb = std::fread(ps_.data(), 1, length_, fp);
+    std::fclose(fp);
+
+    if (readb != length_) {
+      // free mem before bad return
+      ps_.~padded_string();
```
lemire: I am really uneasy about explicit calls to the destructor. I don't think it is idiomatic.

include/simdjson/padded_string.h
DBJDBJ: we have been through this before :)  C++17 standard says if value init aka '{}' is not used variable/constant is initialized to indeterminate value and if opposite it is "zeroed" to "zero" as appropriate to the type in use.  And that is indeed the case for CL (aka MSVC).  

```cpp
// var is "zeroed" to "zero" for the type T
T var{} ;
```

include/simdjson/padded_string.h
DBJDBJ: removed already

src/jsonioutil.cpp
DBJDBJ: this file is to be removed

include/simdjson/jsonioutil.h
DBJDBJ: already removed

include/simdjson/padded_string.h
DBJDBJ: Not sure ... here hitting head one differences between runtimes and compilers.  Leaving it in.

include/simdjson/padded_string.h
lemire: I know the C++ syntax, but isn't viable_size initialize in the constructor?

I should be able to look at this line and understand why it was changed. As it is, I have no idea why you changed it.

By itself, I do not find it objectionable, but you are making many, many changes (hundreds of lines) and I am struggling to find the motivation for most of these changes. This makes it hard to review.

Please change as few lines as possible. 

Can you reorg. your PRs so that they are narrow and focused, with clearly stated benefits?

include/simdjson/padded_string.h
```diff
@@ -17,6 +17,10 @@
 typedef int errno_t ;
 #endif // _MSC_VER
 
+#ifndef SIMDJSON_PADDING
```
lemire: SIMDJSON_PADDING is defined in common_defs.h. Instead of redefinining it again, why not just do

#include "simdjson/common_defs.h"

?

include/simdjson/padded_string.h
```diff
@@ -46,34 +50,37 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
-  bool reset(size_t new_size_) noexcept 
+  static bool reset(type & pstring_, size_t new_size_) noexcept 
   {
-    this->~padded_string(); 
-    viable_size = new_size_;
-    data_ptr = allocate_padded_buffer(new_size_);
-    if (data_ptr != nullptr) {
-      data_ptr[new_size_] = '\0'; // easier when you need a c_str
+      if (pstring_.data())
+        pstring_.~type();
```
lemire: I don't think calling the destructor on a live object is idiomatic.

src/jsonstream.cpp
```diff
@@ -1,9 +1,11 @@
 #include <map>
+#include <stdexcept>
+
 #include "simdjson/jsonstream.h"
 #include "simdjson/isadetection.h"
 #include "jsoncharutils.h"
 
-
+
```
lemire: Please avoid unnecessary white-space changes.

include/simdjson/padded_string.h
```diff
@@ -33,77 +50,160 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
+       a.swap(b) ;
+         printf("%s", a.data() ); // "B"
+  */
+  void swap( padded_string &right_) {
+    std::swap(*this, right_);
   }
 
+  /*
+     padded_string a("A");
+     a.~padded_string() ;
```
lemire: Calling the destructor directly is almost surely not idiomatic.

include/simdjson/padded_string.h
```diff
@@ -33,77 +50,160 @@ inline char *allocate_padded_buffer(size_t length) noexcept {
...
+  explicit padded_string(const char *data, size_t length) noexcept
+      : padded_string(length) 
+  {
     if ((data != nullptr) and (data_ptr != nullptr)) {
       memcpy(data_ptr, data, length);
-      data_ptr[length] = '\0'; // easier when you need a c_str
+      // probably not necessary
+      data_ptr[length] = '\0'; 
+    } else {
+      /* do what? */
```
lemire: The C++ thing to do would be to throw an exception. In our case, I recommend simply assuming that data_ptr might be nullptr and unallocated. I suggest setting the viable_length to zero.

DBJDBJ: OK

include/simdjson/padded_string.h
DBJDBJ: hopefully, no users will do it ...

src/jsonstream.cpp
DBJDBJ: OK

include/simdjson/padded_string.h
DBJDBJ: AFAK I am not doing it in this case. Advice: std::lib code is full of destructor calls.

include/simdjson/padded_string.h
DBJDBJ: ok will remove in next PR

include/simdjson/padded_string.h
DBJDBJ: I am aware of that and will change that behaviour, so we do not indulge in a ping-pong around each PR :) I might do the issue first, discuss it and then do the PR on the agreed outcome of the issue discussion.

src/jsonstream.cpp
DBJDBJ: done

DBJDBJ:
@lemire major mem leak is found and closed ... can we merge this one now?

lemire:
I am not aware of a memory leak in simdjson. Can you elaborate? Is this in relation with an issue what has been submited?

lemire:
As it is, I have concerns with this PR. There are too many changes and it is not clear where the fix is and what the problem was.

DBJDBJ:
> I am not aware of a memory leak in simdjson. Can you elaborate? Is this in relation to an issue that has been submitted?

yesterday I posted (also with a big image in the comment) ... have you not seen that?

image

Here is the fix once more:

```cpp
// padded_string  
// move assignment
  padded_string &operator=(padded_string && other_ ) {
    using namespace std;
    std::swap(data_ptr, other_.data_ptr);
    std::swap(viable_size, other_.viable_size);
    // free the payload
    // remove this and you will have a mem leak
    simdjson::aligned_free_char(other_.data_ptr);
    other_.data_ptr = nullptr;
    other_.viable_size = 0;
    return *this;
  }
```
So, for example ... Each time  `get_corpus()` was returning a padded_stream  instance that instance data was not freed ...

I assume you can create a branch to check for this.

lemire:
Thank you.

DBJDBJ:
> Thank you.

Happy to help. 

After this PR workflow will be: 1. create the issue, 2: implement what was decided in the issue chat.

lemire:
Please see https://github.com/lemire/simdjson/pull/461

DBJDBJ:
Summary of changes 

1. benchmark/get_corpus_benchmark.cpp -- not critical

2. extra/dumpbits.h  -- not critical (it seems that is not used anywhere?)

3. include/simdjson/jsonioutil.h  -- just moved the inline padded_string get_corpus(std::string_view filename) in here, so that jsonioutil.cpp can be removed

4. include/simdjson/padded_string.h -- no mem leak + optimized implementation  
    1. static errno_t load(type  &ps_, FILE *fp) noexcept ; is not used and it thus removed by the compiler, it looks useful, at lease to me

5. src/jsonioutil.cpp -- can be removed

6. src/jsonstream.cpp --  #include <stdexcept> was required

Let me know about fixes required for the failing CI checks.

lemire:
I am going to close this. We have a memory leak and it needs to be fixed. There is another PR for this.

The rest of the changes appear to refer to style and/or refactoring. 

You refer to your other changes to padded_string as "optimizations". If this means "performance enhancement", then please provide benchmark results. If the "optimization" is about "code style that you prefer", then I am not going to merge it.

We are way behind with respect to our move toward version 0.3 and contributions that do not fix bugs or do not get us closer to the next release are discouraged/blocked.

DBJDBJ:
You will get the required padded_string source in the comment in that PR.

PR #477, opened 2020-02-02
Abstract stage 2 tape writing into JsonWriter interface

jkeiser:
This implements #439, decouples stage 2 validation and parsing from tape / iterator production using a [ParsedJsonWriter class](https://github.com/lemire/simdjson/pull/477/files#diff-2a53eba94e6e7a8ab34b425ca397055bR6) with the following interface:

```c++
struct ParsedJsonWriter {
  ErrorValues on_error(ErrorValues error_code);
  ErrorValues on_success(ErrorValues success_code);

  bool on_start_document(uint32_t depth);
  bool on_end_document(uint32_t depth);
  bool on_start_object(uint32_t depth);
  bool on_end_object(uint32_t depth);
  bool on_start_array(uint32_t depth);
  bool on_end_array(uint32_t depth);

  bool on_true_atom();
  bool on_false_atom();
  bool on_null_atom();

  // Returns the destination address where the parser should put parsed strings
  uint8_t *on_start_string();
  // Takes in the final destination pointer so if the parsed json wants, it can calculate length
  bool on_end_string(uint8_t *dst);

  bool on_number_s64(int64_t value);
  bool on_number_u64(uint64_t value);
  bool on_number_double(double value);
};
```

I did this in a series of commits so I could isolate performance dips. `on_start/end_string` actually increases performance somewhat. I'll squash them into one before merging, as it's really not useful in pieces.

This should support all the various tape experiments we have going on. It isn't flexible enough to support the crazy stage 2 decomposition experiment, but I don't think we should optimize for that--we'll solve it as it comes.

This does not eliminate `ParsedJson` references from the structural parser, and that will be needed before we can truly make new tapes. Some things, like `ret_address`, which are really part of the structural validator algorithm--no matter what tape we use, we'll still have that--so should be moved elsewhere. Some things, like structural_indexes, only exist as input from stage 1.

This differs from #463 a bit in naming, and in that it creates a separate `ParsedJsonWriter` class (which I'd like to move more code to!). I'm not particular about the naming. I kind of want to see `ParsedJson` only expose user-facing methods (or at least, I'd like to present such a class to the user even if it's not `ParsedJson`), which is why I made the writer class. It also seems like a natural place to move `current_loc` and `current_string_buffer_loc`, but my first attempts to do so led to performance degradation (see the [visitor-members](https://github.com/lemire/simdjson/tree/jkeiser/visitor-members) branch), so I've left them in ParsedJson for now.

## Performance Learnings

A few performance things I noted during this:

* switch case goto is sensitive to changes

  One of the modifications that can kill perf in stage 2 is using a dynamic goto like `goto pj.ret_address[depth]` directly from inside the switch case. You'll note that even though I moved most of the other logic out of `scope_end` and [into the actual `}` and `]` branches](https://github.com/lemire/simdjson/pull/477/files#diff-8d96b39e2ae1dadf431878c5407518d3R338-R339), we still jump to the label and then [immediately jump away](https://github.com/lemire/simdjson/pull/477/files#diff-8d96b39e2ae1dadf431878c5407518d3R362-R363). Doing the jump away directly in the switch led to a significant (> 10%) performance hit for stage 2.

  My hypothesis is that g++ is generating a static `goto jump_table[c]` for the switch, and stops doing that when you mix in something dynamic.

* Sub-switches can be better

  I attempted to make true, false and null separate arms of the parse_value switch, and it made things a little *slower.* My hypothesis is that atoms are so rare that the decoding cost of more instructions outweighs the speedup for individual atoms. Which probably implies that switches should be ordered in descending order of likelihood ... but I don't know how the decoder and branch predictor interact. Might not actually be a factor.

  It's also possible this is related to the previous point; all three atoms jump to the same place, so it could be that increasing by 2 cases crossed some threshold that increased the switch/goto cost.

* Templatizing interferes with optimizations

  Templatizing `parse_number`--even without making any other change--takes a performance hit. The [templatizing commit](https://github.com/lemire/simdjson/pull/477/commits/f0d3f86d44b8d6d89926375890d2f1ff870b1994) makes stage 2 3% slower than the previous commit on my machine.

  I'd love to know why this is. I'd rather go with the template solution than the current "rely on people 
hooking up the right names to the namespace" solution for implementation switching (the one we use for utf8).

* unified_machine continues to resist refactoring

  I made another desperate attempt to unify the normal parser and the streaming parser, because I'm tired of making the same change in both places ... but performance drops happened. The frustrating bit is that literally the only differences are three if statements outside the main loop: one at the beginning (in start()) and two at the end (in finish()). (And the fact that next_json starts at non-zero and gets updated at the end.) start and finish shouldn't take much of the parse time, yet the drop is significant (I think I got it down to a 5% loss on my best attempt).

  This time my attempt was to make a templatized `structural_parser<bool streaming>` with if (streaming) in the two places that matter. I even went and factored out those bits, leaving structural_parser un-templatized, and having a `unified_machine<bool streaming>(structural_parser parser)` do the different bits). In both cases, performance of the main parser dropped--IIRC, the second one was the better attempt at 5%.


jkeiser:
@lemite I don't believe the use of the `ParsedJsonWriter` intermediary has anything to do with a performance loss: if I hang the `on_*` methods off `ParsedJson`, as you do in #463, performance does not recover: the [jkeiser/visitor2](https://github.com/lemire/simdjson/tree/jkeiser/visitor2) branch does that.

jkeiser:
I'd like to have more discussion, and maybe we'll make more changes, but I'm comfortable with removing the templatization for this checkin and adding a stable way to pick a visitor in a future patch.

jkeiser:
I'm also not opposed to fixing up #463 instead ... I forgot that existed when I sat down to do this, sorry for repeating it :/

jkeiser:
ARM issues are fixed.

lemire:
Looks brilliant.

jkeiser:
Just updated with a clean rebase (I had a duplicate of the commit we just merged from master). Code is source-identical except some dates in the amalgamated source.

lemire:
I see a very small regression, but well under 1%, so it is fine.

jkeiser:
Weird! I actually saw a very small speedup. I'll check this on the Skylakes and ampere.

I'd like to bisect and find out which commit caused the regression. I'm also ok with a tiny regression to get in a maintainable state, but it's worth at least seeing what we can do first.

Though I guess we can merge and do the investigation in parallel.

Do let me know if you'd prefer different naming. I built the patch on my naming intuitions having forgotten you had already made a proposed patch.

lemire:
@jkeiser Please don't investigate.

But if you must...

All around, for input files with numbers, we see the number of instructions slightly up. And it can lead to higher times...



Here are the results on ARM Skylark...

Before this PR:

```
$ ./parsemaster jsonexamples/numbers.json
number of iterations 200

jsonexamples/numbers.json
=========================
     2345 blocks -     150124 bytes - 20003 structurals ( 13.3 %)
special blocks with: utf8         0 (  0.0 %) - escape         0 (  0.0 %) - 0 structurals         0 (  0.0 %) - 1+ structurals      2346 (100.0 %) - 8+ structurals      2345 (100.0 %) - 16+ structurals         0 (  0.0 %)
special block flips: utf8         0 (  0.0 %) - escape         0 (  0.0 %) - 0 structurals         0 (  0.0 %) - 1+ structurals         0 (  0.0 %) - 8+ structurals         1 (  0.0 %) - 16+ structurals         0 (  0.0 %)

All Stages
|    Speed        : 210.5311 ns per block ( 45.16%) -   3.2900 ns per byte -  24.6916 ns per structural -    0.304 GB/s
|    Cycles       : 632.9565 per block    ( 98.19%) -   9.8913 per byte    -  74.2347 per structural    -    3.006 GHz est. frequency
|    Instructions : 790.2945 per block    (100.00%) -  12.3500 per byte    -  92.6876 per structural    -    1.249 per cycle
|    Misses       :    3192 branch misses ( 99.52%) - 6027 cache misses ( 99.99%) - 272497.00 cache references
|- Stage 1
|    Speed        :  83.1100 ns per block ( 17.83%) -   1.2988 ns per byte -   9.7473 ns per structural -    0.770 GB/s
|    Cycles       : 249.5959 per block    ( 38.72%) -   3.9005 per byte    -  29.2732 per structural    -    3.003 GHz est. frequency
|    Instructions : 196.6266 per block    ( 24.88%) -   3.0727 per byte    -  23.0608 per structural    -    0.788 per cycle
|    Misses       :     610 branch misses ( 19.02%) - 2372 cache misses ( 39.35%) - 46651.00 cache references
|- Stage 2
|    Speed        : 126.2276 ns per block ( 27.07%) -   1.9726 ns per byte -  14.8043 ns per structural -    0.507 GB/s
|    Cycles       : 379.3129 per block    ( 58.84%) -   5.9276 per byte    -  44.4867 per structural    -    3.005 GHz est. frequency
|    Instructions : 593.0644 per block    ( 75.04%) -   9.2679 per byte    -  69.5560 per structural    -    1.564 per cycle
|    Misses       :    2523 branch misses ( 78.66%) - 3633 cache misses ( 60.27%) - 225181.00 cache references
```

After this PR:
```
$ ./parse jsonexamples/numbers.json
number of iterations 200

jsonexamples/numbers.json
=========================
     2345 blocks -     150124 bytes - 20003 structurals ( 13.3 %)
special blocks with: utf8         0 (  0.0 %) - escape         0 (  0.0 %) - 0 structurals         0 (  0.0 %) - 1+ structurals      2346 (100.0 %) - 8+ structurals      2345 (100.0 %) - 16+ structurals         0 (  0.0 %)
special block flips: utf8         0 (  0.0 %) - escape         0 (  0.0 %) - 0 structurals         0 (  0.0 %) - 1+ structurals         0 (  0.0 %) - 8+ structurals         1 (  0.0 %) - 16+ structurals         0 (  0.0 %)

All Stages
|    Speed        : 216.7822 ns per block ( 46.21%) -   3.3877 ns per byte -  25.4247 ns per structural -    0.295 GB/s
|    Cycles       : 651.9919 per block    ( 99.49%) -  10.1887 per byte    -  76.4672 per structural    -    3.008 GHz est. frequency
|    Instructions : 794.5580 per block    (100.00%) -  12.4166 per byte    -  93.1877 per structural    -    1.219 per cycle
|    Misses       :    3382 branch misses (100.97%) - 6044 cache misses ( 99.95%) - 272816.00 cache references
|- Stage 1
|    Speed        :  83.0814 ns per block ( 17.71%) -   1.2983 ns per byte -   9.7440 ns per structural -    0.770 GB/s
|    Cycles       : 249.5482 per block    ( 38.08%) -   3.8997 per byte    -  29.2676 per structural    -    3.004 GHz est. frequency
|    Instructions : 196.6266 per block    ( 24.75%) -   3.0727 per byte    -  23.0608 per structural    -    0.788 per cycle
|    Misses       :     617 branch misses ( 18.42%) - 2373 cache misses ( 39.24%) - 46627.00 cache references
|- Stage 2
|    Speed        : 132.9241 ns per block ( 28.33%) -   2.0772 ns per byte -  15.5897 ns per structural -    0.481 GB/s
|    Cycles       : 399.4753 per block    ( 60.96%) -   6.2426 per byte    -  46.8514 per structural    -    3.005 GHz est. frequency
|    Instructions : 597.3278 per block    ( 75.18%) -   9.3345 per byte    -  70.0560 per structural    -    1.495 per cycle
|    Misses       :    2697 branch misses ( 80.52%) - 3651 cache misses ( 60.37%) - 225513.00 cache references
```

So in stage 2, we go from 379.3129 cycles per block to  399.4753 cycles  per block (so a 5% increase). This translates into a 3% regression overall. But it is a small core.

On x64, I was only able to see sub-1% regressions.

It is fine. Don't worry.

lemire:
> Which probably implies that switches should be ordered in descending order of likelihood

The compiler has two ways to implement a switch/case (two main ways). One is as a series of compare/jump. The other one is to use a jump table. I guess it can mix and match.

If you have a jump table, then the order of entries is irrelevant.

If you have many entries in the switch case, you better hope that your switch case is not implemented as a series of compare/jump if the last entries are frequent, for obvious reasons.

lemire:
> I'm also not opposed to fixing up #463 instead ... I forgot that existed when I sat down to do this, sorry for repeating it :/

Do not apologize. I throw away most of my code, often by design.

jkeiser:
Sure, I won't make investigation a prereq to merge.

But that's a big enough jump I'd like to at least see what's up! The nice thing is I can just go to the patch that added the number visitor (assuming that's where the regression is) and try a few different things to see if I can bring it back down.

jkeiser:
> Do not apologize. I throw away most of my code, often by design.

Me too. It's more about respecting another person's work and incorporating their thoughts, for me, than the code itself.

lemire:
> It's more about respecting another person's work

I think you will have to work really hard to insult me by coding around me.

lemire:
> Which probably implies that switches should be ordered in descending order of likelihood

Interestingly, it does not appear that compile abides by the order you provide the cases in!

lemire:
Same code, different compiler, different order!!!

Screen Shot 2020-02-04 at 12 41 43 PM

jkeiser:
My results show speedups (which I frankly have trouble explaining ...) on numbers.json:

| Computer | Master Stage 2 | PR Stage 2 | Speedup |
|---|---|---|---|
| skylake | 1.434 GB/s | **1.469 GB/s** | +2.5% |
| skylake-x | 0.930 GB/s | **0.997 GB/s** | +7.2% |
| 

Methodology:

```bash
git checkout master; git pull; make parse; ./parse -n 10000 jsonexamples/numbers.json; ./parse -n 10000 jsonexamples/numbers.json
git checkout jkeiser/visitor; git pull; make parse; ./parse -n 10000 jsonexamples/numbers.json; ./parse -n 10000 jsonexamples/numbers.json
```

I ran each twice to make sure the number was stable (that's a habit these days).

jkeiser:
There's a lot of variation on ampere today, which I don't normally see ... dockerd seems to be doing some work, maybe I'm conflicting with you?

lemire:
> One of the modifications that can kill perf in stage 2 is using a dynamic goto like goto pj.ret_address[depth] directly from inside the switch case. You'll note that even though I moved most of the other logic out of scope_end and into the actual } and ] branches, we still jump to the label and then immediately jump away. Doing the jump away directly in the switch led to a significant (> 10%) performance hit for stage 2. My hypothesis is that g++ is generating a static goto jump_table[c] for the switch, and stops doing that when you mix in something dynamic.

I'd be surprised if it got compiled to a jump table given that it is only a switch between two states. I cannot see a good reason not to do a sequence of two branches.

Also...

Why would you not expect the compiler to compile this...

```C
...
        case '}':
            goto final;
..
    final:
        goto *table[x];
```
to roughly the same as 


```C
...
        case '}':
             goto *table[x];

```


Why would the compiler jump and then jump again? It seems like a basic optimization to avoid the unnecessary jump.



Now compiling with LLVM clang 6.0, I see no performance difference. So I think we are hitting the quirks of the compiler... not something fundamental.

jkeiser:
I think I'm going to leave it alone as suggested. I did run the numbers, and what I found is strange:

The break point for ARM on numbers.json appears to be [the string visitor](https://g
ithub.com/lemire/simdjson/pull/477/commits/7a200dddfb05e8e04c8ea8ec92accc152c875bfd) patch. Number parsing is unchanged until two commits later. Given that this file *has* no strings, I am befuddled by this result.

Note: I don't trust these numbers 100%, but I did run a crapton. Out of a million iterations or so on various commits after the string patch (and over 500K iterations on the string patch itself) I was unable to get a number higher than 5.29 GB/s. Commits before that (including the one immediately before) were easily able to get 5.36-5.39 GB/s within 100K iterations or so. Weirder still, twitter.json, which has a lot of strings, is basically unaffected!

This means a patch for *strings* caused a 1% overall performance drop on files that only have *numbers.*

The good news, I suppose, is that this particular commit has a lot of possible variations that could be tried. It's definitely a little different from the code we had before (hence the speedup on Intel). I'm spooked, but I don't think I'm going to continue. I want to change where/how strings get parsed anyway.

lemire:
@jkeiser My only criticism has been expressed as a PR:
https://github.com/lemire/simdjson/pull/483

jkeiser:
> I'd be surprised if it got compiled to a jump table given that it is only a switch between two states. I cannot see a good reason not to do a sequence of two branches.

Yeah, two branches should be enough (it's 3 states including error). I don't know what pushes the compiler to decide between branches and jump tables. Note that my hypothesis about the jump table is unverified in assembly because I was tossing out ideas based on the change that fixed the problem, and just kinda moved on to make the next commit, having fixed the issue.

> Also...
> 
> Why would you not expect the compiler to compile this...
> 
> ```C
> ...
>         case '}':
>             goto final;
> ...
>     final:
>         goto *table[x];
> ```
> 
> to roughly the same as 
> 
> ```C
> ...
>         case '}':
>              goto *table[x];
> 
> ```

Well, I understand how a series of statically known addresses can compile to a jump table:

```c++
  switch(c) {
    case '"': goto A;
    case '}': goto B;
    default: goto C;
  }
B:
  goto ret_address;

<compiles to jump table>
static const void* jump_table = { ... &&B, ... }; 
goto jump_table[c];
C:
   goto ret_address;
```

I could totally see this nixing the optimization:

```c++
  switch(c) {
    case '"': goto A;
    case '}': goto B;
    default: goto ret_address;
  }
<jump table>
static const void* jump_table = { ... <??? can't put ret_address here, it's not const static> ... }
```

> Now compiling with LLVM clang 6.0, I see no performance difference. So I think we are hitting the quirks of the compiler... not something fundamental.

Oh, I agree with that for sure. It's just that this explains at least some of the performance drops that have vexed me when trying to change stage 2 :)

lemire:
>  I don't know what pushes the compiler to decide between branches and jump tables.

Possibly some simple heuristics. A table is simple and fast, but it has non-trivial latency... like... 4 cycles or so plus a few cycles to adjust the range and do the validation... and it is still subject to mispredictions... meanwhile, quite a few predicted branches, with instruction fusion, can easily fit inside 4 cycles.

Of course, you also don't want the table to be too large. So that's another consideration.

My guess is that the compiler takes into account the range of values and the number of cases...

lemire:
I am almost certain that this won't get compiled as a jump table:

```C
 switch(c) {
    case '"': goto A;
    case '}': goto B;
    default: goto ret_address;
  }
```

lemire:
Ok. So let us see how many jmp we have currently...

```
$ grep jmp stage2_build_tape.s|wc
    603    1206    7363
```

So 603...

Let me remove all of the goto scope_end...

```
$ grep jmp stage2_build_tape.s|wc
    617    1234    7506
```

So 617...

So 14 more jumps. I think we have two stage 2 (sse and avx), so that's 7 more jumps per stage 2.

Please read carefully: by removing the goto scope_end, I created more jmp in the assembly.

It is the opposite of what you might expect.

Looking at the assembly, it does look like it sometimes "inline" the "goto scope_end" (it does the dynamic goto right there without a prior jump) and sometimes it does not (it does some work and then jump to a location where the jump happens). So two times it will inline and two times it does not inline.

The switch at object_continue and array_continue have an "inlined" goto (there is no intermediate jump). However the jumps at object_begin and array_begin are to an intermediate location where the actual jump happens.


Oddly enough, by replacing just one goto scope_end, I can get the number of jmp even higher (to 620).

Take into account that we have one gigantic function with lots and lots of jumps... like hundreds of them... 

For now I conclude that it is hard to reason about this code at this level.

lemire:
Merge at will.

lemire:
> The switch at object_continue and array_continue have an "inlined" goto (there is no intermediate jump). However the jumps at object_begin and array_begin are to an intermediate location where the actual jump happens.

Notice how convenient this is... The branches are at object_begin and array_begin are never taken (only for empty arrays and objects). We don't care about them. The branches at object_continue and array_continue are commonly taken and they are inlined.

jkeiser:
That does work out nicely :)

jkeiser:
> Let me remove all of the goto scope_end...

Did you remove them or replace them with the dynamic jump from scope_end itself?

> Please read carefully: by removing the goto scope_end, I created more jmp in the assembly.

I actually sort of expect that based on the hypothesis: if it's jumptabling when scope_end is in the picture, and you remove scope_end or replace it with the dynamic jump, it'll probably stop jumptabling and switch to branches. Thus more jmps.

I totally get your point that it ought to already be doing branches, but it's still consistent with that hypothesis. Anyway, it'll be assembly-readin' time next time. This will come up again :)

lemire:
When I wrote that I removed the gotos, I mean that I replaced them with the dynamic jump.

I did not check how the switch were compiled per se. It is not hard, but I assumed that they are sequences of branches.

PR #491, opened 2020-02-11
single header contains local includes in some CI's only

DBJDBJ:
wanted to sort out local includes inside "simdjson" with "simdjson/" in front .. then I noticed some CI's are complaining ... see the image below with the apparent cause...

### DIFF
--- a/include/simdjson/common_defs.h
+++ b/include/simdjson/common_defs.h
@@ -2,7 +2,7 @@
 #define SIMDJSON_COMMON_DEFS_H
 
 #include <cassert>
-#include "simdjson/portability.h"
+#include "portability.h"
 
 // we support documents up to 4GB
 #define SIMDJSON_MAXSIZE_BYTES 0xFFFFFFFF
--- a/include/simdjson/jsonioutil.h
+++ b/include/simdjson/jsonioutil.h
@@ -8,8 +8,8 @@
 #include <stdexcept>
 #include <string>
 
-#include "simdjson/common_defs.h"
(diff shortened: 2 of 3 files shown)


DBJDBJ:
image

jkeiser:
The singleheader version is generated by `make amalgamate`, which you can run locally. I changed it recently to find headers more dynamically, and perhaps that broke something?

I see both `#include "simdjson/x.h"` and `#include "x.h"` in the simdjson.h headers. You're right that we should really pick one.

DBJDBJ:
In any case single header simdjson.h  including "simdjson/x.h " is not a good idea, I might be sure you would agree :) 

> I see both `#include "simdjson/x.h"` and `#include "x.h"` in the simdjson.h headers. You're right that we should really pick one.

If "x.h" and "y.h" are in the same folder, why not just include "x.h" :)

In the repo right now, single header simdjson.h does not contain local includes.  only those very few CI scripts have a problem ...

lemire:
@DBJDBJ @jkeiser 

I am going to close this PR without merging. It fails our tests. If there is an issue with the headers, please submit a patch that passes our tests. Currently, I cannot reproduce any error with the amalgamation script. If there is an error, please suggest a test that exposes the bug.

DBJDBJ:
@lemire what happens if you do this same simple change on your side?

lemire:
The amalgamation script works and is part of the CI. If I make your proposed change then it breaks the build as you report. So we won’t break the build.

PR #510, opened 2020-02-27
Document navigation API

jkeiser:
This enables the following scenarios, tested in `make run_basictests`:

### Array Iteration

```c++
string json(R"( [ 1, 2, 3 ] )");
document doc = document::parse(json);
for (uint64_t value : document::array(doc)) {
  cout << value << endl;
}
```

### Object iteration

```c++
string json(R"( { "a": 1, "b": 2, "c": 3 } )");
document doc = document::parse(json);
for (auto [key, value] : document::object(doc)) {
  // key is a string_view, value can be converted to whatever: int64_t(value) or (const char*)value, for example
}
```

### Error code handling

It has the usual `auto [value, error] = ...` semantics for all calls. (Iteration cannot fail, though the case to `document::array` / `document::object` can.) `as_object()`, `as_array()`, etc. are the methods to call instead of conversion.)

### Documentation

A lot more stuff has doxygen style documentation now.

## TODO

- [X] object iteration
- [X] array iteration
- [X] Scalar string_view, const char*, bool, uint64_t, int64_t, double, null extraction
- [X] error code and exception interfaces
- [X] object["key"] indexing.
- [x] Decide whether and how to protect the user from destroying `document` while iterators are in use. (Do not do this until getting a benchmark up, as most solutions have potential performance impact.)
- [x] README update
- [x] Update important samples in codebase
- [X] Performance testing / benchmarking (including benchmark vs. current document_iterator)
- [x] Decide what to do about document_iterator
- [x] Benchmarking vs. other parsers (does not block this PR)

### DIFF
--- a/tests/basictests.cpp
+++ b/tests/basictests.cpp
@@ -7,10 +7,16 @@
 #include <string>
 #include <vector>
 #include <cmath>
+#include <set>
+#include <string_view>
 
 #include "simdjson/jsonparser.h"
 #include "simdjson/jsonstream.h"
 #include "simdjson/document.h"
+
+#ifndef JSON_TEST_PATH
+#define JSON_TEST_PATH "jsonexamples/twitter.json"
+#endif
 
 // ulp distance
 // Marc B. Reynolds, 2016-2019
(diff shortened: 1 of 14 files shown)


lemire:
Very sexy. This looks like magic!!! At least in the simple cases where we just have an array or an object... we have a few cases like this in our benchmark, but most documents are a tad more complex.

It looks like this PR fails all over the place in CI.

jkeiser:
Yeah, I broke json stream when I moved a bunch of inline functions. I
expect it's a typo.

jkeiser:
I'll make it a point to convert some of the more complex documents, but I'm pretty confident once object indexing is working pretty much all use cases will look like the same magic.

I also need to give error code versions of these examples.

jkeiser:
It's worth noting one thing that is pointedly *not* on the Todo: array size and indexing. Iteration is the only way to interact with arrays, because I don't want our API to give the impression that . I feel like it's ok because the 95% use case for arrays is iteration. Random access is unusual, and when you do it I think you want to convert to a real array first.

I'm uncomfortable with adding object key indexing for the same reasons, but I'm more ok with it because (a) it is so important any API without it would absolutely suck, and (b) most objects don't have hundreds of keys, so the usual use case won't be super terrible.

I will clock performance, though. It will be worth comparing end to end performance to other parsers besides simdjson (we already do, it's just worth expanding).

jkeiser:
Actually, I bet writing a converter to array/vec/map would help a lot with that case. Make it explicit at least.

On that note, I feel handicapped without fluent iteration apis the user can rely on. @dbjdbj @lemire does c++ have the equivalent of <iterable/iterator>.select()/map()/first()/etc.? I've been searching and can't find any. Or do you pretty much have to bring in a non-std library?

lemire:
@jkeiser 

I am not sure what you mean exactly... in standard C++ we have things like  std::transform, std::reduce, std::accumulate...

jkeiser:
That sounds like what I'm looking for, thanks :) I guess I needed to search for the word "algorithms."

lemire:
The idea of relying on implicit conversions done by C++ is ingenious.

lemire:
@jkeiser Visual Studio is still unhappy.

jkeiser:
Note this patch as-is has potential to degrade perf because I did an experiment, moving all inline definitions out of the document and document::parser class definitions, and put them into inline/really_inline definitions in `inline/document.h`. I wanted to see what it looks like if we actually separate interface and implementation, even if the implementation is inline.

On advice from stackoverflow, I removed the word `inline` from the in-class method declarations, and I didn't worry about the order in which the inline functions were declared. Both of those decisions could come back to bite perf in the ass :) But both are easily fixable.

jkeiser:
OK, windows is fixed. String indexing is just about there, and performance checks are next.

lemire:
Do we expect that something like this might be possible...


```C
string json(R"( { "a": 1, "b": 2, "c": 3 } )");
document doc = document::parse(json);
int64_t intval;
string_view key;
for (std::tie(key,intval) : document::object(doc)) {
....
}
```

Or here is another wild thing...


```C
string json(R"( { "a": [1,2], "b": [2,4], "c": [4,3] } )");
document doc = document::parse(json);
int64_t intval;
string_view key;
for (auto [key, value] : document::object(doc)) {
  for(int64_t x : document::array(value) {
  }
}
```

(These are not feature requests, just food for thoughts.)

lemire:
And what about this...

```C
string json(R"( [ 1.1, 2, 3.3 ] )");
document doc = document::parse(json);
for (double value : document::array(doc)) {
  cout << value << endl;
}
```

(again, not a feature request, just food for thoughts)

jkeiser:
> Do we expect that something like this might be possible...
> 
> ```c
> string json(R"( { "a": 1, "b": 2, "c": 3 } )");
> document doc = document::parse(json);
> int64_t intval;
> string_view key;
> for (std::tie(key,intval) : document::object(doc)) {
> ....
> }
> ```

Yeah, I'd like to be able to do this; but we need to separate it into two things:

* First, C++ chokes on `for (std::tie(key,intval) : ...)` :(
* But if I replace it with this, we also presently fail:
  > for ( : document::object(doc)) {
  >   int64_t intval;
  >   string_view key;
  >   std::tie(key,intval) = 
  >   ....
  > }
  Because of the fact that tie destructuring doesn't appear to autoconvert. I suspect we can fix this with `element_result::operator tuple<string_view, int64_t>`, etc. It would be nice ergonomics.
* Iterators have the same problem if you're not using exceptions: `std::tie(intval, error) = *iterator` gives you a standard element, even if you know the type you want for x, and you have to convert it yourself in a separate instruction.

> Or here is another wild thing...
> 
> ```c
> string json(R"( { "a": [1,2], "b": [2,4], "c": [4,3] } )");
> document doc = document::parse(json);
> int64_t intval;
> string_view key;
> for (auto [key, value] : document::object(doc)) {
>   for(int64_t x : document::array(value) {
>   }
> }
> ```

This works already :) You can iterate over any element by casting it to an array.

> And what about this...
> 
> ```c
> string json(R"( [ 1.1, 2, 3.3 ] )");
> document doc = document::parse(json);
> for (double value : document::array(doc)) {
>   cout << value << endl;
> }
> ```
> 
> (again, not a feature request, just food for thoughts)

This works already too :) We upcast uint64_t to double. (Note: we don't presently check for overflow in the conversion, but we have error codes for it. I don't know how I feel about it; I'd rather not take the hit for the check given 99% of cases really never get to this level.)

We just don't downcast double to uint64_t, because I feel like a parser should strive to present lossless information, and *then* the user can explicitly cast away what they don't want.

jkeiser:
Performance appears to be mercifully unaffected on my machine; I was super worried about the mass inline move, but turns out I didn't need to be!

Object indexing is in. I've tested that things like these work:

```c++
document doc = document::parse(R"({ "x": 1, "obj": { "a": 2, "b" 3 } })");
uint64_t x = doc["x"]; // document indexing
uint64_t obj_a = doc["obj"]["a"]; // compound indexing
document::object obj = doc["obj"];
uint64_t a = obj["a"]; // object indexing
uint64_t b = obj["b"];
```

With that, I'm going to work on the Twitter examples I posted in the [API discussion](https://github.com/lemire/simdjson/issues/495#issuecomment-590567742), see if this proves out.

lemire:
Wow.

jkeiser:
OK, I checked in 3 twitter examples to basictests, covering this:

### Basic field access

```c++
// Prints the number of results in twitter.json
document doc = document::parse(get_corpus("jsonexamples/twitter.json"));
uint64_t result_count = doc["search_metadata"]["count"];
cout << "twitter.json has " << result_count << " results" << endl;
```

### Print users with a default profile

```c++
document doc = document::parse(get_corpus("jsonexamples/twitter.json"));
for (document::object tweet : document::array(doc["statuses"])) {
  document::object user = tweet["user"];
  if (user["default_profile"]) {
    string_view name = user["screen_name"];
    cout << "User with no profile: " << string_view(user["screen_name"]) << endl;
  }
}
```

### Aggregate all unique image sizes from tweets

```c++
    // Print image names and sizes
    set<tuple<uint64_t, uint64_t>> image_sizes;
    document doc = document::parse(get_corpus("jsonexamples/twitter.json"));
    for (document::object tweet : document::array(doc["statuses"])) {
      auto [media, not_found] = tweet["entities"]["media"];
      if (!not_found) {
        for (document::object image : document::array(media)) {
          for (auto [key, size] : document::object(image["sizes"])) {
            image_sizes.insert({ size["w"], size["h"] });
          }
        }
      }
    }
```

## Note

Comparing, I think as a user I'd like to use the `as_xxx()` functions when I plan to use a type inline rather than just assigning it:

```c++
// "document::array" gets in my way when reading this
for (document::object tweet : document::array(doc["statuses"])) { ... }
// I like this because the "statuses" is the important part to me, I want to see it as quickly as possible
for (document::object tweet : doc["statuses"].as_array()) { ... }
```

The syntax I like doesn't quite work yet, but that's easily fixable by putting `begin()` and `end()` into `element_result<array>` and `element_result<object>`. I'mma do that.

jkeiser:
as_array / as_object is done, and makes it a lot nicer. [Behold](https://github.com/lemire/simdjson/pull/510/commits/fd0798890b7ca8cbd737d9eddf1ffb3e37ae2a98#diff-d9c0814875fc522bbcf4e70bf78e4d74)!

jkeiser:
OK. I've added iterator benchmarks. The new iterator is generally faster on top of being smaller. For the most complex iterator (the one that balances unique sizes) it is about the same.

| Benchmark | New | Old |
|---|---|---|
| twitter_count | 149 ns | 162 ns |
| twitter_default_profile | 93218 ns | 98639 ns |
| twitter_image_sizes | 47542 ns | 47619 ns |

The new code is significantly easier than the old, as you can see in the benchmarks: the [new version](https://github.com/lemire/simdjson/pull/510/commits/ffbb187347732fdd665a3c19f322e9c01da92fdb#diff-a76a1f91e29a18018a5b0911385340e1R97-R106) of twitter_image_sizes is 10 lines, where the [old iterator version](https://github.com/lemire/simdjson/pull/510/commits/ffbb187347732fdd665a3c19f322e9c01da92fdb#diff-a76a1f91e29a18018a5b0911385340e1R97-R106) is 57 lines. Anecdotally, the old iterator version took me a lot more time to debug, too :)

It's not a super fair comparison, as I know `document_iterator` wasn't designed as a finished user API but was more an exploration of iterator performance. But it's what I have to compare to, so :)

jkeiser:
clang appears to have some issue with sanitize that gcc doesn't (I'm guessing maybe it is trying to dereference an invalid iterator, but that's just a guess) ... I'll hopefully get time later this weekend to try to figure that out, I believe this is baked otherwise.

jkeiser:
Ah, now I understand why we sometimes memcpy stuff that we could have just as easily dereferenced :) It was mad that I was trying to read an unaligned `uint32_t` from the string buffer with a pointer dereference.

jkeiser:
@lemire this one is pretty baked. There are further things to think about (among them protecting the user against deleting the document, and whether to deprecate document_iterator). But those can be thought about in a next PR, I think.

lemire:
cc @croteaucarine

jkeiser:
I've checked off the other TODOs at the top here, as they are covered by other issues and don't particularly block this.

tests/basictests.cpp
```diff
@@ -569,32 +575,233 @@ bool skyprophet_test() {
...
-  if(!bad_example())
-    return EXIT_FAILURE;
-  if(!number_test_powers_of_two())
-    return EXIT_FAILURE;
-  if(!number_test_powers_of_ten())
-    return EXIT_FAILURE;
-  if (!navigate_test())
-    return EXIT_FAILURE;
-  if (!skyprophet_test())
+  // if(!json_issue467())
```
lemire: There is a bunch of tests here that are commented out which is a bit troubling...? If the tests are no longer necessary, they should be removed... but if they are relevant, then they should run.

lemire:
@jkeiser Can you check the commented out tests in basictests? Presumably, you disabled them for a reason?

lemire:
Here is what my static analyzer tells me about document.h... Maybe we want to take these issues into account...

```
Checking include/simdjson/document.h ...
include/simdjson/document.h:40:8: style:inconclusive: Technically the member function 'simdjson::document::is_initialized' can be const. [functionConst]
  bool is_initialized() {
       ^
include/simdjson/document.h:83:3: style:inconclusive: Technically the member function 'simdjson::doc_result::operatorbool' can be const. [functionConst]
  operator bool() noexcept { return error == SUCCESS; }
  ^
include/simdjson/document.h:109:3: style:inconclusive: Technically the member function 'simdjson::doc_ref_result::operatorbool' can be const. [functionConst]
  operator bool() noexcept { return error == SUCCESS; }
  ^
include/simdjson/document.h:164:10: style:inconclusive: Technically the member function 'simdjson::parser::capacity' can be const. [functionConst]
  size_t capacity() { return _capacity; }
         ^
include/simdjson/document.h:169:10: style:inconclusive: Technically the member function 'simdjson::parser::max_depth' can be const. [functionConst]
  size_t max_depth() { return _max_depth; }
         ^
include/simdjson/document.h:77:3: style: Class 'doc_result' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
  doc_result(document &&_doc) : doc(std::move(_doc)), error(SUCCESS) { }
  ^
include/simdjson/document.h:78:3: style: Class 'doc_result' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]
  doc_result(error_code _error) : doc(), error(_error) { }
  ^

```

lemire:
>  It was mad that I was trying to read an unaligned uint32_t from the string buffer with a pointer dereference.


To be fair, the memcpy thing is really not pretty. One could easily create a function `to_uint32_t` or something... However, thankfully, we do not have to do it frequently.

tests/basictests.cpp
jkeiser: Oh hell! I do that sometimes when I have a test failure, and forgot to change it back. I'll fix it.

jkeiser:
I'll look through the static analyses too.

tests/basictests.cpp
jkeiser: (To be clear, I comment out the successful tests in those cases, not the failing ones :))

lemire: > I comment out the successful tests in those cases, not the failing ones 

Where is the fun in that?

jkeiser:
@lemire the first 3 static analysis issues seem like they're from a different revision--they are correct in the version I see. I fixed the next two (thanks). The third and fourth were super intentional, though ...

jkeiser:
OK, assuming this passes tests I'll merge (just uncommented tests and added const to a couple of functions). Thanks for reviewing and catching issues!

lemire:
> assuming this passes tests I'll merge 

Please do.

jkeiser:
This one feels really good :)

lemire:
@jkeiser It is beautiful work, I must say.

PR #537, opened 2020-03-05
Make error_code a class with error_message() method

jkeiser:
- Give error_code an error_message() method: `error.error_message()`
- Enable `cout << error`
- Make it impossible to create error codes with invalid values or manipulate them as integers (since they aren't relatable to one another generally)
- Keep backwards compatibility (deprecated) with people comparing `int` error codes (returned from `json_parse`) with `error_code` objects, and `error_message(int)`
- Rename invalid_json to simdjson_error to reflect its expanded purpose (errors include MEMALLOC as well as navigation errors like NO_SUCH_FIELD).


jkeiser:
This fixes #520.

jkeiser:
Having done this one, I am having doubts. There's a lot of machinery here for basically two small improvements (`error.error_message()` and `cout << error`). `cout << error` can be implemented against `error_code`, handling most of the cases I care about; and `error_message(error)` isn't so bad when you're calling it against C functions like `printf()` anyway.

It provides some safety around error messages, but it's probably not worth the complexity cost. I'll mull it over.

jkeiser:
@lemire I've decided I'm going to abandon this line. We'll keep error codes as enum/int. The abstraction doesn't buy us enough IMO, and it could (for example) make it harder to generate bindings for simdjson. I *would* like to make examples with error codes look better, but I think there's a faster way (`cout << error`). I'll work that up soon.

PR #538, opened 2020-03-06
Emit error message on cerr << error_code

jkeiser:
This helps makes sample code with errors a bit easier on the eyes.

- `cerr << error` now emits the error message.
- `error_message(error_code)` now returns a `const char *` since its primary use is in printf and C functions.
- Because error_code lives in the simdjson namespace, you can now omit the `simdjson::` prefix when calling error_message() because of C++'s namespace lookup.
- Code that uses the `json_parse` (which returns an `int`) is unaffected. `simdjson::error_message(int)` returns `const string &` as before.
- Added missing error message for DEPTH_ERROR
- basictests now validates that all error codes are associated with error messages (and that they are paired with the correct message)


jkeiser:
The readme didn't match my regex, but once I've made the next few changes I'm going to make a readme / tutorial with all the new happiness.

PR #635, opened 2020-03-27
Deprecate error/valid methods on parser

jkeiser:
We're trying to move to a world where the parser doesn't store error information anymore, it's returned from parse and forgotten, so that there's only one place errors exist. To do that, we have to deprecate these methods. (They won't do what you want unless you're using json_parse and build_parsed_json anyway--we only set them from those methods--but deprecating them makes it harder to accidentally do it.)

This patch also fixes any makes the competitions compile without warnings. At this point `make everything` compiles with no warnings!

### DIFF
--- a/benchmark/distinctuseridcompetition.cpp
+++ b/benchmark/distinctuseridcompetition.cpp
@@ -4,6 +4,9 @@
 #include <vector>
 
 #include "benchmark.h"
+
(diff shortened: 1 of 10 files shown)


jkeiser:
This fixes #633.

jkeiser:
Ugh, I need to check where I'm merging more.

PR #666, opened 2020-04-01
Update basics.md

Mengesh:
Fix links to singleheader .h and .cpp


jkeiser:
Fascinating, I wouldn't have expected that to work!

PR #689, opened 2020-04-09
Add most CI tests to cmake

jkeiser:
This replicates `make quicktests` and `make slowtest` from our main Makefile, as another chess move towards removing Makefile (to reduce down to a single build system).

This doesn't quite *finish* quicktests and slowtests: in particular, C++ 11 and 14 tests and tests that deliberately disable haswell and westmere. Those will require a solution that lets you compile against simdjson without using the default options. But may as well move this bit forward since it works and adds a new test (testjson2json.sh).

To run just quicktests, you do:

```bash
cmake .
make
ctest -L quicktests
```


jkeiser:
Yeah. It's one of the last few big bits of tech debt I wanted to clear. I hate build systems, but I tend to work on the things I hate so that I don't have to do them anymore ever.

PR #728, opened 2020-04-17
Compile fuzzers without exceptions, run as part of tests

jkeiser:
This:
- Converts fuzzers to not use exceptions
- Makes fuzz/CMakeLists.txt use the idioms we're using elsewhere, with a link_library to include flags and such
- Makes fuzz/CMakeLists.txt options self-contained
- Makes it so that ctest or ctest -L fuzz will run the fuzzers

I don't actually know *what* the tests are doing, but I can validate that they are in fact running the fuzz functions at least once, because at one point I forgot to check for errors and the fuzzers segfaulted :)


pauldreik:
Will have a look later this week!

pauldreik:
Late to the party, glad to see this merged!

jkeiser:
@pauldreik I'd still love if you can give a once-over. I'm not actually sure what the fuzzers are *doing* when they run as part of the tests, for example :)

PR #737, opened 2020-04-19
Run sanitize using gcc 9

jkeiser:
Turns out gcc 9 sanitize is literally twice as fast as gcc 7, clocking in at [7 minutes vs. 14 minutes](https://app.circleci.com/pipelines/github/simdjson/simdjson/2243/workflows/c31b6fd4-6b47-4b6b-8185-14c2a63265ce). At this point gcc 7 sanitize takes almost as long as all the other CircleCI jobs combined. Is there any reason we wouldn't want to move to it for sanitize (keeping gcc 7 for the other checks)? I'd imagine it also might do a more thorough job finding bugs, but I don't know enough to say that confidently :)

This adds one commit to PR #734 .


jkeiser:
I'll stop making build PRs now, this is my merge limit.

jkeiser:
@lemire are you good with the changes that come from #734 too (part of this PR as well)?

jkeiser:
@lemire we have a lot of code flying around so I'm going to merge this along with #734 ... if you haven't taken a look at that (which you probably already did while looking at this), please do, so I can fix any concerns *post hoc.*

PR #740, opened 2020-04-19
Make amalgamated functionally equivalent to src/simdjson.cpp

pps83:
1) **Amalgamate src/simdjson.cpp as-is**

amalgamation.sh shouldn't change contents of src/simdjson.cpp by forcing dmalloc.h that didn't exist in non-amalgamated version and shouldn't change order of includes by placing simdjson.h at the top

2) **Do not modify amalgamation_demo.cpp from amalgamation.sh**

and update amalgamated headers

### DIFF
--- a/singleheader/amalgamate.sh
+++ b/singleheader/amalgamate.sh
@@ -46,7 +46,7 @@ function doinclude()
         if [[ ! " ${found_includes[@]} " =~ " ${file} " ]]; then
             found_includes+=("$file")
             dofile $AMALGAMATE_INCLUDE_PATH $file
-        fi;
+        fi
     elif [ -f $AMALGAMATE_SOURCE_PATH/$file ]; then
         # generic includes are included multiple times
         if [[ "${file}" == *'generic/'*'.h' ]]; then
@@ -73,10 +73,15 @@ function dofile()
     do
         if [[ "${line}" == '#include "'*'"'* ]]; then
             file=$(echo $line| cut -d'"' -f 2)
+            # include all from simdjson.cpp except simdjson.h
+            if [ "${file}" == "simdjson.h" ] && [ "${2}" == "simdjson.cpp" ]; then
+                echo "$line"
+                continue
+            fi
 
             if [[ "${file}" == '../'* ]]; then
                 file=$(echo $file| cut -d'/' -f 2-)
-            fi;
+            fi
 
             # we explicitly include simdjson headers, one time each (unless they are generic, in which case multiple times is fine)
             doinclude $file $line
@@ -108,15 +113,6 @@ echo "/* auto-generated on ${timestamp}. Do not edit! */" > ${AMAL_H}
 echo "Creating ${AMAL_C}..."
 echo "/* auto-generated on ${timestamp}. Do not edit! */" > ${AMAL_C}
 {
-    echo "#include \"simdjson.h\""
-
-    echo ""
-    echo "/* used for http://dmalloc.com/ Dmalloc - Debug Malloc Library */"
-    echo "#ifdef DMALLOC"
-    echo "#include \"dmalloc.h\""
-    echo "#endif"
-    echo ""
-
     for file in ${ALLCFILES}; do
         dofile $AMALGAMATE_SOURCE_PATH $file
     done
--- a/src/simdjson.cpp
+++ b/src/simdjson.cpp
@@ -1,4 +1,10 @@
 #include "simdjson.h"
+
+/* used for http://dmalloc.com/ Dmalloc - Debug Malloc Library */
+#ifdef DMALLOC
+#include "dmalloc.h"
+#endif
+
 #include "error.cpp"
 #include "implementation.cpp"
 #include "stage1_find_marks.cpp"


pps83:
as a side note, if anybody checks out with crlf and tries to run  amalgamation.sh from [msys2](https://www.msys2.org/), the result will be quite messed up

lemire:
>amalgamation.sh shouldn't change contents of src/simdjson.cpp by forcing dmalloc.h that didn't exist in non-amalgamated version and shouldn't change order of includes by placing simdjson.h at the top

That's feels sensible. The amalgamation script should change as little as possible from the input code.

> Do not modify amalgamation_demo.cpp from amalgamation.sh

Why not?

The rationale for the current setup is that everything in the singleheader repository is a generated artefact. With your proposal, we would have a mix of files which are automatically generated and files that are hand-written.

This mix of automatically generated and hand-coded files in the same directory is something you feel is desirable?

amalgamation.sh
```diff
@@ -73,10 +73,15 @@ function dofile()
     do
         if [[ "${line}" == '#include "'*'"'* ]]; then
             file=$(echo $line| cut -d'"' -f 2)
+            # inlcude all from src/${AMAL_C} except ${AMAL_H}
```
jkeiser, line 76: typo inlcude -> include

amalgamation.sh
```diff
@@ -73,10 +73,15 @@ function dofile()
...
+            if [ "${file}" == "${AMAL_H}" ] && [ "${RELFILE}" == "src/${AMAL_C}" ]; then
```
jkeiser, line 77: Oh, this is so that #include "simdjson.h" happens at the beginning of simdjson.cpp, in all cases?

How did anything ever work?? I assume it's because each of the .cpp files includes their own header files?

If that's the case, perhaps the right thing to do is remove #include "simdjson.h" from the top of simdjson.cpp?

Either way, this is a super good catch!

amalgamation.sh
```diff
@@ -100,80 +105,29 @@ echo "/* auto-generated on ${timestamp}. Do not edit! */" > "${AMAL_H}"
...
 echo "Try :"
 echo "c++ -O3 -std=c++17 -pthread -o ${CPPBIN} ${DEMOCPP}  && ./${CPPBIN} ../jsonexamples/twitter.json ../jsonexamples/amazon_cellphones.ndjson"
 
-SINGLEHDR=$SCRIPTPATH/singleheader
-echo "Copying files to $SCRIPTPATH/singleheader "
+echo "Copying files to $SINGLEHDR"
 mkdir -p $SINGLEHDR
 echo "c++ -O3 -std=c++17 -pthread -o ${CPPBIN} ${DEMOCPP}  && ./${CPPBIN} ../jsonexamples/twitter.json ../jsonexamples/amazon_cellphones.ndjson" > $SINGLEHDR/README.md
-cp ${AMAL_C} ${AMAL_H}  ${DEMOCPP} $SINGLEHDR
+cp ${AMAL_C} ${AMAL_H} $SINGLEHDR
```
jkeiser, line 130: I'm actually super OK with this, I just blithely kept the copy out of momentum. @lemire are we OK with not putting amalgamation_demo into the produced library anymore?

We should remove the amalgamation_demo.cpp from singleheader as well, if so.

lemire: > We should remove the amalgamation_demo.cpp from singleheader as well, if so.


I don't understand. This makes it so that the script does not create the demo program. If we do not include it, how are we going to run checks on the single-header code? Of course, we could compile some other piece of code.

lemire:
Ok. I think that the issue to resolve is whether we want to continue having the script generate amalgamation_demo.cpp or not.

The options are...

1. We keep things as they are. If it ain't broken, don't fix it.

2. We remove amalgamation_demo.cpp entirely. We do not really need it. Instructions need to change, but that's a small thing.

3. We keep amalgamation_demo.cpp in some form. The current form won't do... have a look at it, it says that it is automatically generated.


I think that what does this PR (which we want to merge) is not quite correct. It does not fit into 1, 2, 3. Rather it is ...

4. Keep amalgamation_demo.cpp as it is, as a file that is labeled as automatically generated.



I am not exactly sure why "1" is not the best option here. 

Again: I openly discourage changes without identifiable benefits. We don't want to change things on a whim.

amalgamation.sh
pps83: The demo program is always constant. script shouldn't generate it from the start. In this PR there are two commits, one removes that generation of amalgamation_demo.cpp, perhaps, it should later be moved to examples or some other folder out of singleheader dir

amalgamation.sh
pps83: So, the reason it all works: 
amalgamation first builds simdjson.h, and adds all the headers to the list of processed headers, and simdjson.cpp is built, all these header from list of processed are skipped including simdjson.h. As, I've mentioned in ["Do not inline file IO functions"]("https://github.com/simdjson/simdjson/pull/741) PR, I wanted to add `_CRT_SECURE_NO_WARNINGS`, which must be at the top of simdjson.cpp before any other headers, and because of the way amalgamation was done, it forces simdjson.h to be at the top. This PR addresses it, and makes sure that amalgamated simdjson.cpp is identical to the original.

pps83:
> 2. We remove amalgamation_demo.cpp entirely. We do not really need it. Instructions need to change, but that's a small thing.

IMO, amalgamated or not, it shouldn't matter to the users. Results must be identical. That is, any other demo should work effectively the same well regardless if the source was amalgamated or not.
The only difference, is that it include simdjson.cpp to avoid compiling simdjson.cpp separately. Test one-liner cmd to build demo could instead list two cpp files to generate test executable.

jkeiser:
@lemire note that you can presently do this to run *all* tests against the amalgamated headers:

```
SIMDJSON_TEST_AMALGAMATED_HEADERS=1 make test
```

I'd be sympathetic to a change that removed amalgamation_demo and added the above to CI, for that reason.

One concrete reason amalgamation_demo sometimes causes me issues is when I change the base DOM API, add deprecations, etc., the amalgamation script doesn't show up when I do a "Find In Files" .cpp / .h. Having all our C++ code in .cpp and .h files is nice for editors. It's not big enough that I'd feel like making a fuss about it, though.

The primary reason we'd *want* amalgamation_demo.cpp is if we wanted to *ship* it with the header files as a sort of quickstart. I'm on board with that!

lemire:
Note that this PR looks to be obsolete given @jkeiser's recent work.

pps83:
> Note that this PR looks to be obsolete given @jkeiser's recent work.

In what way? The problem was not addressed in any way, lots of changes were done to conflict with the PR, and all these changes likely were happening after the PR was started. The main point of the PR: "Make amalgamated functionally equivalent to src/simdjson.cpp" - this problem exists, was not taken care of in any way. Why other changes to move it, and related to amalgamation_demo.cpp were done before taking a fix to amalgamation.sh?

The point of PRs  and contributions, as I see it, is to accept them, or ask for changes and accept them. Not to start doing the same/similar/conflicting work on raised subjects.

jkeiser:
The relevant PR was done 5 days ago and just took a while to review. It doesn't attempt to address the problems herein.

jkeiser:
#729.

Sorry about the conflict.

jkeiser:
So. I think we can find a way to *remove* amalgamate_demo.cpp now, if you're into that. We could compile quickstart with it instead of amalgamate_demo.cpp. That sidesteps any questions of whether all the files should be generated or not, and resolves the concern that unchanging source should be in .cpp files in the repository.

@pps83 I still don't understand the dmalloc.h thing. Why are we adding that? I don't see it anywhere else in the repository, so it doesn't seem to be a parity thing to make amalgamation the same as the src/ files?

lemire:
1. I do not give a damn about the DMALLOC thing. This was added because someone on Twitter was insistant that simdjson was broken since it could not work with custom memory allocators.

2. I don't care about  amalgamate_demo.cpp. Burn it.

pps83:
> 1. I do not give a damn about the DMALLOC thing. This was added because someone on Twitter was insistant that simdjson was broken since it could not work with custom memory allocators.
> 2. I don't care about  amalgamate_demo.cpp. Burn it.

I agree on both.

DMALLOC most likely won't even work. simdjson.h itself doesn't use it. I guess that someone had no clue what they were doing. When I saw it, I was wondering how possibly it could work if it only affects simdjson.cpp. Perhaps, they had to add dmalloc always before the headers and in all impl files, otherwise it would try to allocate/deallocate with different allocators :)

amalgamate_demo shouldn't be specific to amalgamate. Regular demo should work the same with alamgamated or not version of simdjson.

jkeiser:
My impression of where we are:

- Keeping #include "simdjson.h" in simdjson.cpp is a positive change.
- Moving #include "dmalloc.h" to src/simdjson.cpp is necessary for compatibility, Thanks! I'm worried about removing it but also not opposed, if both of you think it's fine. Not for this, but if we support DMALLOC we should test it :) I honestly hadn't noticed it until now.
- Let's not move amalgamation_demo.cpp into the source tree with this change. The right change is to eliminate it altogether and compile quickstart against the amalgamated headers, but we shouldn't feel like we have to hold up this patch until it's done.

lemire:
@jkeiser  +1

pps83:
I updated only amalgamate.sh to preserve simdjson.cpp as-is and moved dmalloc to simdjson.cpp. This way amalgamated and regular are now identical.

jkeiser:
:+1:. Will give @lemire a chance to look at the final state before merging.

lemire:
If @jkeiser  approves, it is good enough.

pps83:
@jkeiser FYI, some of the changes caused this:

image

amalgamated version used to have paths, now only names. `end file` message doesn't even show what file ends. I think it was better before (on the right in the screenshot is the way it used to be before). Note, this has nothing to do with this current PR

lemire:
@pps83 This does look like a regression that we want to address. Can you open an issue?

jkeiser:
Oh, sorry, that must have been a stealth change. I changed it to show src or include prefix so that I could see where in our tree it came from. Happy to change it back though.

jkeiser:
It was in with the changes when I made amalgamate work for Windows.

lemire:
It might be nice to quickly determine the matching source file... useful for debugging.

jkeiser:
Happiness :)

PR #743, opened 2020-04-20
Master vs2019 x86 compile fixes

pps83:
1) **Added bitexact implementations of _BitScanForward64 and _BitScanReverse64 for VS2019 32-bit builds**

2) **Added bitexact implementations of _umul128 for VS2019 x86, arm, arm64 builds**

3) **Implement mul_overflow for VS2019 arm64 builds**

4) **Added Win32 for VS2019 to .appveyor.yml**

and updated amalgamated headers


jkeiser:
@lemire could you have a look at this one? I'm less qualified on the ins and outs of the intrinsics than you :)

lemire:
Thank you.

This actually looks good.

singleheader/simdjson.cpp
```diff
@@ -5771,10 +5770,31 @@ struct value128 {
   uint64_t high;
 };
 
+#if defined(_MSC_VER) && !defined(_M_X64) // _umul128 for x86, arm, arm64
+#if defined(_M_ARM)
+static inline uint64_t __emulu(uint32_t x, uint32_t y) {
+  return x * (uint64_t)y;
+}
+#endif
+static inline uint64_t _umul128(uint64_t ab, uint64_t cd, uint64_t *hi) {
```
lemire, line 5779: For ARM 64-bits, I think that this should be _umul128 instead of this expensive 32-bit emulation.

https://github.com/miranda-ng/miranda-ng/blob/9aa5329a4b4077970571f40f79486a3a1bf3e9f8/libs/libmdbx/src/test/utils.h#L53

lemire: Oh. I see that you are ahead of me. Good.

singleheader/simdjson.cpp
```diff
@@ -5771,10 +5770,31 @@ struct value128 {
...
+  uint64_t bd = __emulu((uint32_t)ab, (uint32_t)cd);
+  uint64_t adbc = ad + __emulu((uint32_t)ab, (uint32_t)(cd >> 32));
+  uint64_t adbc_carry = !!(adbc < ad);
+  uint64_t lo = bd + (adbc << 32);
+  *hi = __emulu((uint32_t)(ab >> 32), (uint32_t)(cd >> 32)) + (adbc >> 32) + (adbc_carry << 32) + !!(lo < bd);
+  return lo;
+}
+#if defined(_M_ARM64)
+uint64_t umul128(uint64_t ab, uint64_t cd, uint64_t* hi) { // non-inline extern version for arm64
+  return _umul128(ab, cd, hi);
```
lemire, line 5790: Good.

singleheader/simdjson.cpp
lemire: Oh. Wait... No.

singleheader/simdjson.cpp
lemire: Maybe not so good. You are calling the 32-bit emulated version even if you have _umul128? https://github.com/miranda-ng/miranda-ng/blob/9aa5329a4b4077970571f40f79486a3a1bf3e9f8/libs/libmdbx/src/test/utils.h#L53

src/jsoncharutils.h
```diff
@@ -324,10 +324,31 @@ struct value128 {
...
+  uint64_t bd = __emulu((uint32_t)ab, (uint32_t)cd);
+  uint64_t adbc = ad + __emulu((uint32_t)ab, (uint32_t)(cd >> 32));
+  uint64_t adbc_carry = !!(adbc < ad);
+  uint64_t lo = bd + (adbc << 32);
+  *hi = __emulu((uint32_t)(ab >> 32), (uint32_t)(cd >> 32)) + (adbc >> 32) + (adbc_carry << 32) + !!(lo < bd);
+  return lo;
+}
+#if defined(_M_ARM64)
+uint64_t umul128(uint64_t ab, uint64_t cd, uint64_t* hi) { // non-inline extern version for arm64
+  return _umul128(ab, cd, hi);
```
lemire, line 344: This would only be acceptable if Visual Studio has no native _umul128. It is maybe not the case, but you can get the high 64-bits of a multiplication using a single instruction with the __umulh intrinsic, see...

 https://stackoverflow.com/questions/28868367/getting-the-high-part-of-64-bit-integer-multiplication

src/arm64/bitmanipulation.h
```diff
@@ -60,14 +60,13 @@ really_inline bool add_overflow(uint64_t value1, uint64_t value2, uint64_t *resu
 }
 
 #ifdef _MSC_VER
-#pragma intrinsic(_umul128) // todo: this might fail under visual studio for ARM
```
lemire, line 63: Microsoft's documentation always include "#pragma intrinsic(_umul128)". It is being removed from the code in this PR. Why? Either it is unnecessary, or else...?

src/jsoncharutils.h
lemire: So I think the proper approach under ARM 64 bits is...
```C++
hi = __umulh(ab, cd);
return ab * cd;
```

src/arm64/bitmanipulation.h
lemire: Note: I realize that the intrinsic may only be available under x64. Still... do we need the pragma?

lemire:
@pps83 


You seem to claim that this makes simdjson build under Visual Studio using the 32-bit mode...

Could you modify the .appveyor.yml file so that it gets tested?

This is purely optional.

singleheader/simdjson.cpp
```diff
@@ -698,14 +698,13 @@ really_inline bool add_overflow(uint64_t value1, uint64_t value2, uint64_t *resu
 }
 
 #ifdef _MSC_VER
-#pragma intrinsic(_umul128) // todo: this might fail under visual studio for ARM
+uint64_t umul128(uint64_t ab, uint64_t cd, uint64_t* hi);
```
lemire, line 701: this should be marked really_inline. We want to be sure that it gets inlined.

lemire:
@pps83 This is nice work. Please see my requests.

singleheader/simdjson.cpp
pps83: in this case the function is in different cpp files, so, no way to inline it as this would work only assuming all cpp files are amalgamated in specific order. In any case, your suggestion for _M_AMR64 is much better, and I'll be able to remove the extern completely.

src/arm64/bitmanipulation.h
pps83: I think your own code should never add this pragma (in this and all other places). You should only include proper headers to use it.

pps83:
> ```c++
> hi = __umulh(ab, cd);
> return ab * cd;
> ```

good suggestion. Code updated.
On a side note, for clz/ctz, modern x86 cpus have an instruction instead of using that bitscanforward/reverse stuff. For VS 64-bit there is [__lzcnt64](https://docs.microsoft.com/en-us/cpp/intrinsics/lzcnt16-lzcnt-lzcnt64?view=vs-2019), it will produce the same code as __builtin_clzll, except that it won't work for x86.

For arm/arm64 with ms compiler, you could use [_CountLeadingZeros64](https://docs.microsoft.com/en-us/previous-versions/windows/embedded/ms913693(v%3Dmsdn.10)):
image

pps83:
> Could you modify the .appveyor.yml file so that it gets tested?

I added a change to `.appveyor.yml`, but I'm not familiar with it, so, not sure if that's OK.

src/arm64/bitmanipulation.h
lemire: Do you have a reference supporting this statement? All I could find online is this...

"The intrinsic pragma tells the compiler that a function has known behavior. The compiler may call the function and not replace the function call with inline instructions, if it will result in better performance."

I'll ask on twitter.

lemire: Note that Microsoft's documentation gives examples where both the header is included and the pragma called https://docs.microsoft.com/en-us/cpp/intrinsics/umul128?view=vs-2019

I expect I will know more soon.

singleheader/simdjson.cpp
```diff
@@ -697,16 +697,10 @@ really_inline bool add_overflow(uint64_t value1, uint64_t value2, uint64_t *resu
...
-#endif
-
 really_inline bool mul_overflow(uint64_t value1, uint64_t value2, uint64_t *result) {
 #ifdef _MSC_VER
-  // todo: this might fail under visual studio for ARM
-  uint64_t high;
-  *result = _umul128(value1, value2, &high);
-  return high;
+  *result = value1 * value2;
+  return !!__umulh(value1, value2);
```
lemire, line 703: This is probably fine on ARM 64, but it is suboptimal on x64. Here is what it compiles to:
```asm
        mov     rax, rcx
        imul    rax, rdx
        mov     QWORD PTR [r8], rax
        mov     rax, rdx
        mul     rcx
        test    rdx, rdx
        setne   al
```
Notice the two multiplications?

If you try, instead the following...
```C++
  uint64_t high; 
  *result = _umul128(value1, value2, &high);		
   return high;
```
You get the much better assembly on x64...

```asm
        mov     rax, rdx
        mul     rcx
        test    rdx, rdx
        mov     QWORD PTR [r8], rax
        setne   al
```

lemire:
> On a side note, for clz/ctz, modern x86 cpus have an instruction instead of using that bitscanforward/reverse stuff. For VS 64-bit there is __lzcnt64, it will produce the same code as __builtin_clzll, except that it won't work for x86.


That is what we use in the haswell kernel, see 
https://github.com/simdjson/simdjson/blob/a8e892ba907b7561a3203315a65e788b98b2e537/src/haswell/bitmanipulation.h#L36

We fall back on _BitScanReverse64 / __builtin_clzll in the westmere kernel.

lemire:
>For arm/arm64 with ms compiler, you could use _CountLeadingZeros64, count_ones should also be changed from neon to regular intrinsic (arm to neon and back would most likely be a huge hit).

I do not think so. Not for modern ARM cores. We have optimizations manuals and we run regular benchmarks on ARM-based servers.

To my knowledge, there is no (non-NEON) popcnt instruction on ARM. The best routine (to my knowledge) is the NEON function you found. Now, if you have a faster way to do a population count, I'd love to have a look and see what it compiles to.

lemire:
> For arm/arm64 with ms compiler, you could use _CountLeadingZeros64,

I agree. Let us handle this as a separate issue.

lemire:
Merging. 

@pps83 If you can share your name, I'd love to add you to our list of contributors. Of course, you will be listed on GitHub via your commit as a contributor.

pps83:
> Merging.
> 
> @pps83 If you can share your name, I'd love to add you to our list of contributors. Of course, you will be listed on GitHub via your commit as a contributor.

Sure, my name is Pavel Pavlov
Thanks

singleheader/simdjson.cpp
pps83: That block is completely in simdjson:: arm64 namespace. Arm-only code for _MSC_VER

src/arm64/bitmanipulation.h
pps83: Yes, that's weird that they put it that way in the examples. AFAIK, intrin.h already does define it as intrinsic, and compiler obviously knows that it's intrinsic

pps83: In any case, pragma thing is irrelevant for the PR: function is changed not to use `_umul128` here, so, the pragma was removed also.

PR #744, opened 2020-04-20
Mark const variables as static

pps83:


lemire:
Can you offer a rationale?

And how do you feel about the following answer...

> At namespace scope a constant already has internal linkage by default, so
adding 'static' is only adding a deprecated feature that does nothing.

Source: https://bytes.com/topic/c/answers/141429-static-const-variables-within-namespace

jkeiser:
In C++, a static variable is allowed to have multiple copies in different translation units. I hadn't realized namespaces did something different, but even if it wasn't in a namespace, I don't think we'd want to add `static` to the singleton implementations. It would imply they were not actually singletons.

For structural_or_whitespace_or_exponent_or_decimal_negated, I could take it or leave it, but it sounds like static will have no effect either way?

lemire:
@jkeiser I think that...

```C++
namespace {
int x;
}
```

gives you internal linkage...

But not, I think...
```C++
namespace simdjson {
int x;
}
```

So the idea is that you do not need static in C++.

lemire:
Anyhow, we need a rationale for this change.

pps83:
> Anyhow, we need a rationale for this change.

My non-scientific rationale: if compiler sees a static variable,  then it knows for sure that nothing outside of current compilation unit can refer to the variable, and, as a result, may do a better job at optimizing it.

lemire:
> My non-scientific rationale: if compiler sees a static variable, then it knows for sure that nothing outside of current compilation unit can refer to the variable, and, as a result, may do a better job at optimizing it.

Ok so your rationale is better performance. I tried your code against master and out of all of our files... you are not saving a single instruction using GNU GCC 8.3. The running times are identical... for example, for twitter.json, we get a cost of 309.4212 instructions per block... very precisely.

If the motivation is better performance, that's great... what is also great is that we are measuring very precisely our performance... so if there is a benefit to a coding strategy, we should see it.

Have you run your own benchmarks... do they agree with my results?


Or do you see a performance gain that you can report?

pps83:
In this case, obviously, it won't have any difference and binary will almost certainly be bit exact.

it's just more correct not to make vars extern if they aren't used outside of the file.

src/implementation.cpp
```diff
@@ -100,7 +100,7 @@ class unsupported_implementation final : public implementation {
   unsupported_implementation() : implementation("unsupported", "Unsupported CPU (no detected SIMD instructions)", 0) {}
 };
 
-const unsupported_implementation unsupported_singleton{};
+static const unsupported_implementation unsupported_singleton{};
```
jkeiser, line 103: I think we want there only to be one of these--therefore it should not be static. With static, the compiler will not warn you if you try to make multiple copies, but with const it will. The reason we want one is we're going to be taking the address of it and I wouldn't mind the option of being able to *compare* the address.

jkeiser: Same goes for the others in this file.

src/generic/numberparsing.h
```diff
@@ -217,7 +217,7 @@ really_inline bool is_integer(char c) {
 // We need to check that the character following a zero is valid. This is
 // probably frequent and it is harder than it looks. We are building all of this
 // just to differentiate between 0x1 (invalid), 0,1 (valid) 0e1 (valid)...
-const bool structural_or_whitespace_or_exponent_or_decimal_negated[256] = {
+static const bool structural_or_whitespace_or_exponent_or_decimal_negated[256] = {
```
jkeiser, line 220: This I can take or leave. It has no bearing either way, since this file is only going to be included once per namespace. const is enough to let the compiler know that the values herein will always be the same as at compile time.

Can it be constexpr? That would tell the compiler what we want it to know more explicitly. I still suspect it doesn't make a difference, but it'd be more accurate, I think?

src/jsoncharutils.h
```diff
@@ -11,7 +11,7 @@ namespace simdjson {
 
 // these are the chars that can follow a true/false/null or number atom
 // and nothing else
-const uint32_t structural_or_whitespace_or_null_negated[256] = {
+static const uint32_t structural_or_whitespace_or_null_negated[256] = {
```
jkeiser, line 14: constexpr question here too. Same comments as numberparsing.h.

src/jsoncharutils.h
```diff
@@ -678,641 +678,641 @@ static const components power_of_ten_components[] = {
...
-                                 0x1858ccfce06cac74,
-                                 0xf37801e0c43ebc8,
-                                 0xd30560258f54e6ba,
-                                 0x47c6b82ef32a2069,
-                                 0x4cdc331d57fa5441,
-                                 0xe0133fe4adf8e952,
-                                 0x58180fddd97723a6,
-                                 0x570f09eaa7ea7648};
-
+const uint64_t mantissa_128[] = {
```
jkeiser, line 681: @lemire I actually like this change for readability. Large indents are jarring. It's entirely possible I made this change when I was being more conservative about style at the beginning of collaboration :)

src/jsoncharutils.h
lemire: Yes, constexpr works, I tried it. I think that's what it should be.

src/generic/numberparsing.h
lemire: Yes, it should be constexpr.

lemire:
I'll issue a distinct PR making these arrays constexpr as I think that's the right qualifier.

src/jsoncharutils.h
lemire: @jkeiser This is the LLVM style as interpreted by clang-format...

https://github.com/simdjson/simdjson/blob/master/.clang-format

We can change that, of course.

jkeiser: Oh, if there is a style formatter already I won't impose; I'd even rather enforce it :)

lemire: We have not been using it much but it is there... 

```
$ ls style/
clang-format-check.sh clang-format.sh       run-clang-format.py
```

lemire: To achieve the reformatting needed, we need to add a trailing comma to the long array prior to formatting...

Screen Shot 2020-04-24 at 5 35 40 PM

lemire:
So I made a distinct PR that achieves much of the same result. https://github.com/simdjson/simdjson/pull/797

I am not excited either by my own PR or by this current PR.

I do not see any concrete benefit. 

I think we should drop this PR as well as my own version (https://github.com/simdjson/simdjson/pull/797) until actual benefits can be identified.

Furthermore, this current PR makes the code more error prone as argued by @jkeiser.

So I recommend shutting this down.

I will restate my general rule: contributions, code changes, should have clearly identifiable benefits. If the benefits are merely hypothetical, it is typically better to wait and see.

pps83:
> I will restate my general rule: contributions, code changes, should have clearly identifiable benefits. If the benefits are merely hypothetical, it is typically better to wait and see.

With all these arrays in simdjson, it won't affect generated code in no way, these are basic data arrays that compiler will optimize in the same way regardless of  static or constexpr. The idea behind static is to ensure that it's not extern symbol. It's not about optimizations.

IMO, when one adds static array it has to be from the start static (and const if it's not to be modified). Only if the array is to be accessible from other cpp files static should be removed.

jkeiser:
I see. I still think constexpr is the right move, but I can also see why generating simdjson as a .so might cause the compiler to make different decisions than if it was compiled into an executable, since it has to assume other people can access it.

lemire:
I could see the idea of reducing the number of exported symbols as something useful to do... (maybe... why would anyone ever care about this? don't we have more pressing worries?) but this PR would not make a dent in our list...

```
$ objdump -t ./libsimdjson.so|grep simdjson
./libsimdjson.so:     file format elf64-x86-64
0000000000000000 l    df *ABS*	0000000000000000              simdjson.cpp
00000000000291a0 l     O .data	0000000000000300              _ZN8simdjson8fallbackL10jump_tableE
0000000000029900 l     O .bss	0000000000000050              _ZN8simdjson8internalL21unsupported_singletonE
0000000000024c20 l     O .rodata	0000000000000400              _ZN8simdjsonL24structural_or_whitespaceE
0000000000020100 l     O .rodata	0000000000000100              _ZN8simdjson8fallback13stringparsing12_GLOBAL__N_1L10escape_mapE
0000000000023e40 l     O .rodata	0000000000000dd8              _ZN8simdjson12_GLOBAL__N_1L14digit_to_val32E
0000000000025020 l     O .rodata	0000000000000400              _ZN8simdjson12_GLOBAL__N_1L32structural_or_whitespace_negatedE
00000000000215e0 l     O .rodata	00000000000027a0              _ZN8simdjson12_GLOBAL__N_1L23power_of_ten_componentsE
0000000000020000 l     O .rodata	0000000000000100              _ZN8simdjson8fallback13numberparsing12_GLOBAL__N_1L55structural_or_whitespace_or_exponent_or_decimal_negatedE
0000000000023d80 l     O .rodata	00000000000000b8              _ZN8simdjson12_GLOBAL__N_1L12power_of_tenE
0000000000020200 l     O .rodata	00000000000013d0              _ZN8simdjson12_GLOBAL__N_1L12mantissa_128E
0000000000020100 l     O .rodata	0000000000000100              _ZN8simdjson7haswell13stringparsing12_GLOBAL__N_1L10escape_mapE
0000000000020000 l     O .rodata	0000000000000100              _ZN8simdjson7haswell13numberparsing12_GLOBAL__N_1L55structural_or_whitespace_or_exponent_or_decimal_negatedE
0000000000020100 l     O .rodata	0000000000000100              _ZN8simdjson8westmere13stringparsing12_GLOBAL__N_1L10escape_mapE
0000000000020000 l     O .rodata	0000000000000100              _ZN8simdjson8westmere13numberparsing12_GLOBAL__N_1L55structural_or_whitespace_or_exponent_or_decimal_negatedE
0000000000025420 l     O .rodata	0000000000000800              _ZN8simdjsonL14thintable_epi8E
0000000000025d40 l     O .rodata	0000000000000100              _ZN8simdjsonL19BitsSetTable256mul2E
0000000000025c20 l     O .rodata	0000000000000110              _ZN8simdjsonL20pshufb_combine_tableE
0000000000004590 l     F .text	0000000000000672              _GLOBAL__sub_I_simdjson.cpp
0000000000029a80 l     O .bss	0000000000000050              _ZN8simdjson8internalL17haswell_singletonE
0000000000029a20 l     O .bss	0000000000000050              _ZN8simdjson8internalL18westmere_singletonE
00000000000299c0 l     O .bss	0000000000000050              _ZN8simdjson8internalL18fallback_singletonE
0000000000029960 l     O .bss	0000000000000050              _ZN8simdjson8internalL59detect_best_supported_implementation_on_first_use_singletonE
00000000000042e0 l     F .text	000000000000017f              _GLOBAL__sub_I_simdjson.cpp.cold.56
00000000000198f0  w    F .text	00000000000000d9              _ZN8simdjson8internal26unsupported_implementationC2Ev
0000000000018f60  w    F .text	0000000000000042              _ZN8simdjson8internal26unsupported_implementationD2Ev
0000000000018e20  w    F .text	0000000000000006              _ZNK8simdjson8internal26unsupported_implementation5parseEPKhmRNS_3dom6parserE
00000000000122b0 g     F .text	00000000000034cf              _ZNK8simdjson8westmere14implementation6stage2EPKhmRNS_3dom6parserE
0000000000028a48  w    O .data.rel.ro	0000000000000010              _ZTIN8simdjson14implementationE
00000000000190a0  w    F .text	0000000000000044              _ZN8simdjson8internal49detect_best_supported_implementation_on_first_useD0Ev
0000000000018e70  w    F .text	0000000000000042              _ZN8simdjson8westmere14implementationD2Ev
000000000000ec50 g     F .text	00000000000035dd              _ZNK8simdjson7haswell14implementation6stage2EPKhmRNS_3dom6parserERm
0000000000028cb0  w    O .data.rel.ro	0000000000000060              _ZTVN8simdjson8westmere14implementationE
0000000000018f10  w    F .text	0000000000000042              _ZN8simdjson8fallback14implementationD1Ev
00000000000191e0  w    F .text	000000000000003e              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use6stage2EPKhmRNS_3dom6parserERm
0000000000018da0 g     F .text	000000000000004b              _ZNK8simdjson8westmere14implementation5parseEPKhmRNS_3dom6parserE
0000000000025f20  w    O .rodata	0000000000000048              _ZTSN8simdjson8internal49detect_best_supported_implementation_on_first_useE
0000000000028b90  w    O .data.rel.ro	0000000000000060              _ZTVN8simdjson8internal26unsupported_implementationE
0000000000019590  w    F .text	00000000000001a0              _ZN8simdjson7haswell13numberparsing19parse_large_integerEPKhRNS_3dom6parserEb
0000000000005130 g     F .text	0000000000000098              _ZNK8simdjson8internal29available_implementation_list21detect_best_supportedEv
0000000000019220  w    F .text	000000000000002c              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use6stage2EPKhmRNS_3dom6parserE
0000000000018f60  w    F .text	0000000000000042              _ZN8simdjson8internal26unsupported_implementationD1Ev
0000000000005100 g     F .text	0000000000000006              _ZNK8simdjson8internal29available_implementation_list4sizeEv
0000000000028ab8  w    O .data.rel.ro	0000000000000018              _ZTIN8simdjson8internal26unsupported_implementationE
0000000000018d10 g     F .text	000000000000004b              _ZNK8simdjson7haswell14implementation5parseEPKhmRNS_3dom6parserE
0000000000018e40  w    F .text	0000000000000006              _ZNK8simdjson8internal26unsupported_implementation6stage1EPKhmRNS_3dom6parserEb
0000000000019250  w    F .text	000000000000003f              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use6stage1EPKhmRNS_3dom6parserEb
0000000000008500 g     F .text	000000000000004b              _ZNK8simdjson8fallback14implementation5parseEPKhmRNS_3dom6parserE
0000000000019320  w    F .text	00000000000000ba              _ZN8simdjson17codepoint_to_utf8EjPh
0000000000028b30  w    O .data.rel.ro	0000000000000060              _ZTVN8simdjson8internal49detect_best_supported_implementation_on_first_useE
00000000000051d0 g     F .text	0000000000000105              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use8set_bestEv
0000000000028a70  w    O .data.rel.ro	0000000000000018              _ZTIN8simdjson8westmere14implementationE
000000000001be80  w    F .text	000000000000075d              _ZN8simdjson8westmere6stage113json_minifier6minifyILm64EEENS_10error_codeEPKhmPhRm
0000000000019140  w    F .text	0000000000000044              _ZN8simdjson7haswell14implementationD0Ev
0000000000018e70  w    F .text	0000000000000042              _ZN8simdjson8westmere14implementationD1Ev
0000000000018ec0  w    F .text	0000000000000042              _ZN8simdjson7haswell14implementationD2Ev
00000000000193e0  w    F .text	00000000000001a3              _ZN8simdjson8fallback13numberparsing19parse_large_integerEPKhRNS_3dom6parserEb
0000000000019730  w    F .text	00000000000001a0              _ZN8simdjson8westmere13numberparsing19parse_large_integerEPKhRNS_3dom6parserEb
0000000000018cd0 g     F .text	0000000000000011              _ZNK8simdjson7haswell14implementation6minifyEPKhmPhRm
0000000000019000  w    F .text	0000000000000044              _ZN8simdjson8fallback14implementationD0Ev
0000000000018e50  w    F .text	0000000000000006              _ZNK8simdjson8internal26unsupported_implementation6stage2EPKhmRNS_3dom6parserE
0000000000025ee0  w    O .rodata	0000000000000025              _ZTSN8simdjson8fallback14implementationE
0000000000028a58  w    O .data.rel.ro	0000000000000018              _ZTIN8simdjson7haswell14implementationE
0000000000025e40  w    O .rodata	000000000000001c              _ZTSN8simdjson14implementationE
0000000000025f80  w    O .rodata	0000000000000031              _ZTSN8simdjson8internal26unsupported_implementationE
0000000000018e60  w    F .text	0000000000000006              _ZNK8simdjson8internal26unsupported_implementation6stage2EPKhmRNS_3dom6parserERm
0000000000018d80 g     F .text	0000000000000012              _ZNK8simdjson8westmere14implementation6stage1EPKhmRNS_3dom6parserEb
00000000000191c0  w    F .text	0000000000000018              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use4nameB5cxx11Ev
0000000000019300  w    F .text	0000000000000019              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use11descriptionB5cxx11Ev
0000000000019050  w    F .text	0000000000000044              _ZN8simdjson8internal26unsupported_implementationD0Ev
0000000000008550 g     F .text	0000000000003232              _ZNK8simdjson8fallback14implementation6stage2EPKhmRNS_3dom6parserERm
00000000000191a0  w    F .text	0000000000000019              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use25required_instruction_setsEv
0000000000004d90 g     F .text	000000000000036a              _ZNK8simdjson8fallback14implementation6stage1EPKhmRNS_3dom6parserEb
00000000000294f0 g     O .bss	0000000000000008              _ZN8simdjson8internal21active_implementationE
00000000000190f0  w    F .text	0000000000000044              _ZN8simdjson8westmere14implementationD0Ev
0000000000005120 g     F .text	0000000000000008              _ZNK8simdjson8internal29available_implementation_list3endEv
00000000000052e0 g     F .text	0000000000000079              _ZN8simdjson8fallback13numberparsing18slow_float_parsingEPKcRNS_3dom6parserE
0000000000018fb0  w    F .text	0000000000000042              _ZN8simdjson8internal49detect_best_supported_implementation_on_first_useD2Ev
0000000000028ad0  w    O .data.rel.ro	0000000000000060              _ZTVN8simdjson14implementationE
0000000000005360 g     F .text	000000000000319a              _ZNK8simdjson8fallback14implementation6stage2EPKhmRNS_3dom6parserE
0000000000029500 g     O .bss	00000000000003e8              _ZN8simdjson8internal11error_codesE
0000000000018e00  w    F .text	0000000000000005              _ZNK8simdjson14implementation11descriptionB5cxx11Ev
0000000000018f10  w    F .text	0000000000000042              _ZN8simdjson8fallback14implementationD2Ev
00000000000198f0  w    F .text	00000000000000d9              _ZN8simdjson8internal26unsupported_implementationC1Ev
0000000000018cf0 g     F .text	0000000000000012              _ZNK8simdjson7haswell14implementation6stage1EPKhmRNS_3dom6parserEb
0000000000018d60 g     F .text	0000000000000011              _ZNK8simdjson8westmere14implementation6minifyEPKhmPhRm
000000000001c5e0  w    F .text	00000000000015da              _ZN8simdjson8westmere6stage123json_structural_indexer5indexILm64EEENS_10error_codeEPKhmRNS_3dom6parserEb
00000000000192d0  w    F .text	000000000000002c              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use5parseEPKhmRNS_3dom6parserE
0000000000018df0  w    F .text	0000000000000005              _ZNK8simdjson14implementation4nameB5cxx11Ev
0000000000012230 g     F .text	0000000000000079              _ZN8simdjson8westmere13numberparsing18slow_float_parsingEPKcRNS_3dom6parserE
0000000000028aa0  w    O .data.rel.ro	0000000000000018              _ZTIN8simdjson8internal49detect_best_supported_implementation_on_first_useE
00000000000294e0 g     O .bss	0000000000000008              _ZN8simdjson21active_implementationE
000000000001a4d0  w    F .text	00000000000019a1              _ZN8simdjson7haswell6stage123json_structural_indexer5indexILm128EEENS_10error_codeEPKhmRNS_3dom6parserEb
0000000000025fc0 u     O .rodata	0000000000000020              _ZZN8simdjson8westmere15utf8_validation13is_incompleteENS0_4simd5simd8IhEEE9max_array
0000000000019290  w    F .text	000000000000003e              _ZNK8simdjson8internal49detect_best_supported_implementation_on_first_use6minifyEPKhmPhRm
00000000000199d0  w    F .text	0000000000000af4              _ZN8simdjson7haswell6stage113json_minifier6minifyILm128EEENS_10error_codeEPKhmPhRm
0000000000018fb0  w    F .text	0000000000000042              _ZN8simdjson8internal49detect_best_supported_implementation_on_first_useD1Ev
0000000000025ea0  w    O .rodata	0000000000000025              _ZTSN8simdjson8westmere14implementationE
0000000000025e60  w    O .rodata	0000000000000024              _ZTSN8simdjson7haswell14implementationE
0000000000025fe0 u     O .rodata	0000000000000020              _ZZN8simdjson7haswell15utf8_validation13is_incompleteENS0_4simd5simd8IhEEE9max_array
0000000000004cd0 g     F .text	0000000000000066              _ZNK8simdjson8fallback14implementation6minifyEPKhmPhRm
00000000000294e8 g     O .bss	0000000000000001              _ZN8simdjson25available_implementationsE
0000000000005110 g     F .text	0000000000000008              _ZNK8simdjson8internal29available_implementation_list5beginEv
0000000000018e30  w    F .text	0000000000000006              _ZNK8simdjson8internal26unsupported_implementation6minifyEPKhmPhRm
0000000000018ec0  w    F .text	0000000000000042              _ZN8simdjson7haswell14implementationD1Ev
0000000000028bf0  w    O .data.rel.ro	0000000000000060              _ZTVN8simdjson8fallback14implementationE
0000000000028a88  w    O .data.rel.ro	0000000000000018              _ZTIN8simdjson8fallback14implementationE
000000000000b810 g     F .text	0000000000003440              _ZNK8simdjson7haswell14implementation6stage2EPKhmRNS_3dom6parserE
0000000000028c50  w    O .data.rel.ro	0000000000000060              _ZTVN8simdjson7haswell14implementationE
0000000000015780 g     F .text	0000000000003548              _ZNK8simdjson8westmere14implementation6stage2EPKhmRNS_3dom6parserERm
0000000000018e10  w    F .text	0000000000000004              _ZNK8simdjson14implementation25required_instruction_setsEv
000000000000b790 g     F .text	0000000000000079              _ZN8simdjson7haswell13numberparsing18slow_float_parsingEPKcRNS_3dom6parserE
```

jkeiser:
Oh, I don't know that I'd go on a crusade, but I maybe learned a new way to think about static :) Still, since these are const, the compiler should be able to lift out any constants it wants at compile time as well.

pps83:
> I see. I still think constexpr is the right move, but I can also see why generating simdjson as a .so might cause the compiler to make different decisions than if it was compiled into an executable, since it has to assume other people can access it.

Yes, afaik. on linux all extern (eg non static) symbols are exported from .so by default

pps83:
By the way, these arrays are part of amalgamated sindjson.h, you might be really shocked/surprised, that a project that uses simdjson.h may end up with 50 copies of all these static array if the project ends up including simdjson.h from different .cpp files. I added a simple sample project that shows the problem, try it: https://github.com/pps83/string-link-demo

lemire:
Can you elaborate on what you mean?

If you flag an array as being static, you do not expect to have just one copy. Right?

pps83:
> Can you elaborate on what you mean?
> 
> If you flag an array as being static, you do not expect to have just one copy. Right?

Yes I do. But it doesn't seem to be happening with MS compiler (not sure about others). That project shows the issue.

jkeiser:
static doesn't mean the same thing outside of classes that it does inside: it means that there can be multiple physical copies. Especially for non-const things. I don't think it's a huge problem for us to risk multiple copies of these arrays, but we should make sure we're buying something real for it :)

lemire:
So let me reiterate this because I think that there is a misunderstanding.

What I am saying is that if you go from "const T x" to "static const T x", you should expect to see multiple copies of the array in the compiled output.

So when I wrote "If you flag an array as being static, you do not expect to have just one copy. Right?"... I meant that adding the "static" qualifier, I would expect to see extra copies popping up.

pps83:
> So let me reiterate this because I think that there is a misunderstanding.
> 
> What I am saying is that if you go from "const T x" to "static const T x", you should expect to see multiple copies of the array in the compiled output.
> 
> So when I wrote "If you flag an array as being static, you do not expect to have just one copy. Right?"... I meant that adding the "static" qualifier, I would expect to see extra copies popping up.

If the static array is in a header, then yes, multiple copies could exist (but linker may merge them, and I'd expect linker to do so).

If the static array is in a single .cpp file, then it doesn't get exposed to linker (and isn't exported from .so)

pps83:
> static doesn't mean the same thing outside of classes that it does inside: it means that there can be multiple physical copies (one per compilation unit). Especially for non-const things. I don't think it's a huge problem for us to risk multiple copies of these arrays, because we only define most of these in a single compilation unit; but we should make sure we're buying something real for it :)

I checked, and it seems that all the data arrays are in simdjson.cpp, so, there is no possibility of duplication. Static is't really needed, the only real difference is that it will hide all these array from exporting in .so

jkeiser:
@pps83 I think we have concern over static vs. constexpr. I don't see a reason not to merge a constexpr change in principle, since it does clarify the intended compile-time meaning of these arrays. I also don't mind terribly if we want to hide these from the .so, but if we go with a private namespace change, I'd like to know whether we'll end up with multiple copies in the executable or not--that's my only concern.

So, where I land on this is, *as is* we should close it. But I could imagine an update to use constexpr for clarity if it avoid's @lemire's noted antipattern, and/or a patch that doesn't expose these in the .so but also ensures there is a single copy of the variables.

pps83:
I think constexpr for these arrays won't change anything, they are used at runtime afaik (but I didn't really look into it). Since simdjson is effectively a single cpp file all of these arrays are going to be in a single cpp file, they aren't in headers, so it doe make sense to make them `static` or `static constexpr`, both would be the same

jkeiser:
@pps83 OK. I'm going to close, but the concern about exposing symbols is noted and real :) Out of curiosity, have you found that exposing the symbols causes some practical issues where people take dependencies on the symbols, or is it more the theoretical concern that if it's exposed, someone *could* write their own header and access the variable?

pps83:
> @pps83 OK. I'm going to close, but the concern about exposing symbols is noted and real :) Out of curiosity, have you found that exposing the symbols causes some practical issues where people take dependencies on the symbols, or is it more the theoretical concern that if it's exposed, someone _could_ write their own header and access the variable?

Not real practical issues imo, it's the same as making private stuff in class private: if you make something public that shouldn't be, it would be strange. Private stuff should be private

lemire:
Note that you can directly control which symbols get exported...

http://anadoxin.org/blog/control-over-symbol-exports-in-gcc.html

pps83:
> Note that you can directly control which symbols get exported...
> 
> http://anadoxin.org/blog/control-over-symbol-exports-in-gcc.html

Yes, if you use hidden by default and then manually export what you want to export (windows/ms style)

PR #753, opened 2020-04-20
Fix cmake error when SIMDJSON_COMPETITION=OFF

jkeiser:
Typo.


jkeiser:
Merging (looks green).

PR #799, opened 2020-04-25
Make checkperf work on Windows

jkeiser:
This:

* Enables command line options on all command line programs on windows via a compatibility getopt.h
* Enables parse -t on Windows
* Moves checkperf into cmake primitives instead of bash
* Enables ctest -R checkperf on Windows

This uses -DSIMDJSON_CHECKPERF_BRANCH=jkeiser/parse-t, which is a branch *just* containing the changes to enable parse -t on Windows. It will be removed on the next checkin (we can't run it against master until it supports parse -t).


jkeiser:
I did this a little differently from how I did the competition dependencies; I made it so it would only pull and build the repository if you actually specify the checkperf target, rather than doing it at configure time. This doesn't seem to be the usual pattern, but I like it better. I hate paying for stuff I don't use.

jkeiser:
@lemire this one also downsizes our Windows CI runs to 3, still checking all individual options but not trying as many combinations.

PR #803, opened 2020-04-25
Run Windows tests and builds in parallel

jkeiser:
This relies on cmake having a version that supports parallel on appveyor, but there's a reasonable chance of that ...

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -24,12 +24,13 @@ build_script:
   - set
   - mkdir build
   - cd build
+  - cmake --version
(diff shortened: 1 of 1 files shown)


jkeiser:
--parallel without arguments did absolutely nothing. Trying --parallel 4, just to see.

jkeiser:
--parallel with no arguments is fine, it just wasn't faster.

@lemire this is *barely* faster, but I guess it's faster (3-4 minutes or so per checkin, a little less than a minute faster per build). Every little bit helps.

PR #804, opened 2020-04-25
Disable deprecation warnings for VS builds

pps83:
fopen/getenv are standard c++ that' are not deprecated.


jkeiser:
I'm super excited for this, it's our last warning! WOOO

pps83:
> I'm super excited for this, it's our last warning! WOOO

not yet. There is one more:

```
C:\simdjson-pps\singleheader>cl /c simdjson.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.25.28614 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

simdjson.cpp
The contents of <string_view> are available only with C++17 or later.
The contents of <optional> are available only with C++17 or later.
C:\\2019\Enterprise\VC\Tools\MSVC\14.25.28610\include\ostream(408): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc
C:\\2019\Enterprise\VC\Tools\MSVC\14.25.28610\include\ostream(401): note: while compiling class template member function 'std::basic_ostream<char,std::char_traits<char>> &std::basic_ostream<char,std::char_traits<char>>::operator <<(double)'
C:\simdjson-pps\singleheader\simdjson.h(4852): note: see reference to function template instantiation 'std::basic_ostream<char,std::char_traits<char>> &std::basic_ostream<char,std::char_traits<char>>::operator <<(double)' being compiled
C:\simdjson-pps\singleheader\simdjson.h(2318): note: see reference to class template instantiation 'std::basic_ostream<char,std::char_traits<char>>' being compiled
```

simply open VS dev console, and try to compile manually

jkeiser:
Weird! Why didn't that one show up before?

Are exception semantics turned off by default with VS2019?

pps83:
> Weird! Why didn't that one show up before?
> 
> Are exception semantics turned off by default with VS2019?

I also find it weird if compiler itself disables exceptions by default :)  No idea

jkeiser:
If that's the case, we either need to turn SIMDJSON_EXCEPTIONS off when /EHsc is not enabled, or disable the warnings. Gotta think about what will actually happen if it stayed on though--I presume it's abort().

pps83:
with `/EHsc` there are no warnings. I think it's ok to ignore it or maybe add code to handle it in place where it triggers that warning

jkeiser:
I filed #807 on the /EHsc thing.

pps83:
PR updated (conflicts resolved)

jkeiser:
@pps83 I should've separated my concerns into separate comments; the part I'm most concerned about is localizing the PUSH/POP to just the individual functions or even lines, if possible) that call fopen/getenv. I'd really like to get the benefit of the warnings in CI so we can *evaluate* violations of this for buffer overruns and the like before we disable. If it gets onerous we can consider another strategy later.

pps83:
> @pps83 I should've separated my concerns into separate comments; the part I'm most concerned about is localizing the PUSH/POP to just the individual functions or even lines, if possible) that call fopen/getenv. I'd really like to get the benefit of the warnings in CI so we can _evaluate_ violations of this for buffer overruns and the like before we disable. If it gets onerous we can consider another strategy later.

I don't see any issues. This PR does two things:
 1) [pragma warning push](https://docs.microsoft.com/en-us/cpp/preprocessor/warning?view=vs-2019) stores warning state and at the end of the file I restore it. This way, including simdjson.h in any other project won't affect warnings in users' code. Note, that `SIMDJSON_PUSH_DISABLE_WARNINGS` obviously doesn't disable warnings in simdjons.h/cpp. The name is wrong, it should have been called `SIMDJSON_PUSH_WARNINGS`/`SIMDJSON_POP_WARNINGS`
 2) disables 4996 warning (for using deprecated functions). This way, even though you don't define `_CRT_SECURE_NO_WARNINGS` ms compiler ignores the fact that simdjson.h uses "deprecated" functions.

and in simdjson.cpp 4996 is entirely disabled. 


I would certainly not try to minimize that warning to functions/lines. The warning is bogus, it's a compiler bug. There is no benefit in having that warning in simdjons's code. By deprecating standard c++ this warning totally looses it's credibility and value: if one constantly sees it in the build output, it becomes routinely ignored.
Possible questionable value simdjson might have from not having that warning enabled: if somebody added `std::auto_ptr` to simdjson.h/cpp then you'd see that warning.

jkeiser:
@pps83 the warning you are disabling is deprecation in general, not _CRT_SECURE_NO_WARNINGS.

jkeiser:
@pps83 the warning you are disabling is deprecation in general, not _CRT_SECURE_WARNINGS. It may be bogus for this purpose, but it's not bogus for all. I'd like to know if we violate it when calling other things.

Per _CRT_SECURE_WARNINGS, I don't know whether to call it a compiler bug, but it's not valueless. It is a warning that you are using functions that are commonly used in dangerous ways. The compiler's advice on what to change is wrong, though.

Anyway, I won't fight about it too much. Your PR makes things better and I'm willing to localize the warnings myself :)

jkeiser:
Merged via #849.

PR #816, opened 2020-04-27
Remove unneeded .gitignores

jkeiser:
Now that cmake is the only build system, we don't generate
binaries in the source tree anymore. Amalgamate also doesn't
leave anything around.


jkeiser:
Tested this by doing:

```
mkdir build
cd build
cmake -DSIMDJSON_GOOGLE_BENCHMARKS=ON ..
make
ctest
```

And then, back from the root, based on fuzz/Fuzzing.md:

```
fuzz/build_like_ossfuz.sh
mkdir -p out/parser
build/fuzz/fuzz_parser out/parser/
fuzz/build_fuzzer_variants.sh
```

PR #851, opened 2020-05-01
Propagate some CMake variables to checkperf

furkanusta:
Although it passes the user-defined options to checkperf, if the project is build in Debug mode or with Clang (since CXX defaults to gcc on Linux) results can fluctuate

I am suspecting that CMAKE_BUILD_TYPE that was missing in #841 might be the reason for simdjson-internal to fail when built as a static library. I was using different directories for shared and static builds and might have forgotten to pass the build type. Will re-check, when I rebase #841  on top of these PRs


benchmark/checkperf.cmake
```diff
@@ -58,7 +58,9 @@ if (Git_FOUND)
   # - second, cmake ..
   add_custom_command(
     OUTPUT ${SIMDJSON_CHECKPERF_DIR}/build/cmake_install.cmake # We make many things but this seems the most cross-platform one we can depend on
-    COMMAND ${CMAKE_COMMAND} -DSIMDJSON_GOOGLE_BENCHMARKS=OFF -DSIMDJSON_COMPETITION=OFF -G ${CMAKE_GENERATOR} ..
+    COMMAND
+    ${CMAKE_COMMAND} -E env CXX=${CMAKE_CXX_COMPILER} CC=${CMAKE_C_COMPILER}
```
jkeiser, line 62: I wanted to point you at the previous step, which relies on the top-level simdjson-user-cmakecache.cmake. This provides an initial CMakeCache.txt populated with the decisions we've made, so that we hopefully compile the same thing when we compile master, and compare apples to apples. The compiler decision could (maybe?) be added to that, but I have a feeling it doesn't work that way for CXX and CC.

I have no issue with the change you've made, and I'm not 100% sure adding CMAKE_C*_COMPILER to CMakeCache.txt would do the right thing, so I'm approving. But I did want to get your thoughts on the pattern. Would add_subdirectory(${SIMDJSON_CHECKPERF_DIR}) even work, and would it be the right thing to do?

furkanusta: I was actually going to ask the same question. I suppose initially you used this mechanism so that checkperf uses the same options, but if I remember correctly add_subdirectory builds with the same variables. I'll test add_subdirectory if it works it would be easier than manually invoking cmake

jkeiser: Thanks! I was just afraid of it since I expected it would share export names and stuff.

PR #852, opened 2020-05-02
CMake: Separate public and private compilation flags

furkanusta:
simdjson-internal-flags for macros and warnings
simdjson-flags for pthread, sanitizer, and libcpp

I kept the name internal since that was the one I had used for simdjson in #841


jkeiser:
I haven't missed looking at this one, I just want to double check callsites more thoroughly to ensure we are using the right flags everywhere.

furkanusta:
By the way, since I will have to rebase #841 on top of these recent commit, I can ditch the -internal suffix and use -public -private (or something else) if it makes things more clear

jkeiser:
> By the way, since I will have to rebase #841 on top of these recent commit, I can ditch the -internal suffix and use -public -private (or something else) if it makes things more clear

I'd leave `simdjson-flags` for public flags--just leave the good names to public things. private would be fine with me, internal is also fine. I wish I had a preference to give you ... well, I suppose the word "private" is used in doxygen and c++ so if we used -private we could have one fewer word in our lexicon. I don't care enough to want anyone to do work, though.

simdjson-flags.cmake
```diff
@@ -35,39 +35,40 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
...
   endif()
-  target_compile_options(simdjson-flags INTERFACE -Werror -Wall -Wextra -Wsign-compare -Wshadow -Wwrite-strings -Wpointer-arith -Winit-self -Wconversion -Wno-sign-conversion)
+  target_compile_options(simdjson-internal-flags INTERFACE -Werror -Wall -Wextra -Wsign-compare -Wshadow -Wwrite-strings -Wpointer-arith -Winit-self -Wconversion -Wno-sign-conversion)
 endif()
 
 # Optional flags
 option(SIMDJSON_IMPLEMENTATION_HASWELL "Include the haswell implementation" ON)
 if(NOT SIMDJSON_IMPLEMENTATION_HASWELL)
-  target_compile_definitions(simdjson-flags INTERFACE SIMDJSON_IMPLEMENTATION_HASWELL=0)
+  target_compile_definitions(simdjson-internal-flags INTERFACE SIMDJSON_IMPLEMENTATION_HASWELL=0)
```
jkeiser, line 53: @lemire @furkanusta I'm trying to decide what to do about these. If you do this:

```cmake
set(SIMDJSON_IMPLEMENTATION_HASWELL=OFF)
add_subdirectory(simdjson) # or find_project? I'm not familiar enough
add_executable(blah blah.cpp)
target_link_libraries(blah simdjson)
```

Then do we expect simdjson to compile without HASWELL in it? ... I think we do.

jkeiser: (Note: example is aspirational and may or may not compile.)

singleheader/CMakeLists.txt
```diff
@@ -94,5 +94,5 @@ add_dependencies(simdjson-singleheader-source amalgamate)
 # Test the generated simdjson.cpp/simdjson.h using the generated amalgamate_demo.cpp
 #
 add_executable(amalgamate_demo $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/amalgamate_demo.cpp>)
-target_link_libraries(amalgamate_demo simdjson-singleheader-include-source simdjson-flags)
+target_link_libraries(amalgamate_demo simdjson-singleheader-include-source simdjson-internal-flags simdjson-flags)
```
jkeiser, line 97: I think we should have: `target_link_libraries(simdjson-internal-flags simdjson-flags)` so we don't have to remember to do this :)

simdjson-flags.cmake
jkeiser: Yeah, this is the only thing I think we need to change--these should be a part of simdjson-flags so they are public interface.

furkanusta: If find_package is used in CMake, then we are using an already-built library and it is not possible to deactivate HASWELL etc.

However, it is possible with add_directory

```CMake
set(SIMDJSON_IMPLEMENTATION_HASWELL ON CACHE STRING "" FORCE)
add_subdirectory(simdjson EXCLUDE_FROM_ALL)
add_executable(foo foo.cpp)
target_link_libraries(foo PUBLIC simdjson)
```
This works even now. It still builds the library but now we can control the parameters.

In this case _HASWELL option in CMake is already part of the public API. Users do not need to now which macros this option enables, and making the macros public felt like we are polluting the consumer's project.

furkanusta:
I've added target_link_libraries(simdjson-internal-flags simdjson-flags) 
One drawback of this is that public flags are linked twice for the library.
we need to link simdjson-internal-flags privately and simdjson-flags publicly, and simdjson-internal-flags also contains simdjson-flags. Thus, all the public parameters are list twice while building the library (consuming is fine). 
None of the flags we are using right now causes problems when repeated, and I am not aware of any others.

jkeiser:
OK. Not a fan of duplicating flags, but forgetting them is worse, ergo this is better :)

simdjson-flags.cmake
jkeiser: Grokked. So the rubric is "any flags that affect the headers should be in `simdjson-flags`, but flags affecting only the .cpp should be in `simdjson-internal` in anticipation users will use the library. Works for me.

As an aside, I really appreciate that you push back on suggestions you think are wrong, it is so easy for people to just grumble and accede to requests just to get a patch in.

PR #867, opened 2020-05-07
[do not merge] test if checkperf can detect a 4-5% degradation

jkeiser:
This degrades overall perf by around 5% on my machine, for whatever reason (from 4350 down to 4100 docs/sec). Let's see if the new checkperf branch catches it.


jkeiser:
It's looking like the answer here is "no," but let's see if the old checkperf can ... creating a PR for that.

jkeiser:
Old checkperf doesn't either, so we don't have a good test yet. We can try running it against the string length patch though.

PR #868, opened 2020-05-07
[do not merge] Move string writing functions to stage 2

jkeiser:
This causes a 5% regression on my machine, and I'm interested to see if CI catches it.


jkeiser:
It doesn't. Dang it.

PR #883, opened 2020-05-14
Make include files more fine-grained

jkeiser:
document.h is rather bulky and hard to work with. This splits document.h and inline/document.h into:
   - dom/array.h, inline/array.h
   - dom/document.h, inline/document.h
   - dom/element.h, inline/element.h
   - dom/object.h, inline/object.h
   - dom/parser.h, inline/parser.h
   - internal/tape_type.h
   - internal/tape_ref.h, inline/tape_ref.h

It also attempts to minimize header dependencies, including only headers necessary to compile a given file. Principally, this means forward declaring classes in header files when the included classes are only used in method declarations.

I also moved `scope_descriptor` into the internal namespace, which I think was an oversight before.

Aside from changing the namespace of scope_descriptor, I made no actual changes to code or definitions here; this is just moving files around, and occasionally adding forward declarations to deal with include file order.

Performance does not appear to have changed.


jkeiser:
I'll add a note to the readme that anything other than simdjson.h is internal-only. I reasoned that if anyone ever *was* including the files directly, they were already in for a world of hurt when we split out inline functions and changed the dependency ordering.

jkeiser:
Churn complete. Had these in the back of my mind for a while, and figured we'd get it all in one swoop when you suggested breaking out the stage 2 files.

PR #887, opened 2020-05-15
Don't repeat circleci checkperf tests on drone

jkeiser:
We run perf tests on both circleci and drone, and they seem to fail spuriously more often on Drone. This removes all non-arm checkperfs from drone. We rely on circleci for those now.

I'm hoping this will remove most of the spurious checkperf failures without losing the huge value of a CI performance check.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,7 @@ steps:
     CC: gcc
     CXX: g++
     BUILD_FLAGS: -- -j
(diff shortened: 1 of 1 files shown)


jkeiser:
> If I were smarter and harder-working, I'd setup up my own CI server. :-)

Me too! We should form a club for people who would do lots more stuff if they were harder working. I nominate someone else to actually do the work to form the club.

PR #972, opened 2020-06-23
Compile with -Weffc++ in CI

jkeiser:
Turns out we haven't been compiling with these warnings on in CI because we have SIMDJSON_GOOGLE_BENCHMARKS enabled. This helps explain why I sometimes get extra warnings and sometimes don't--when I am running with different options!


jkeiser:
Yeah, I believe the if statement there is a relic from when we had global compiler flags.

PR #1012, opened 2020-06-30
Add script I used to explain the lookup4 tables

jkeiser:
It's *really* not something we'll probably run again, but it's related to the project and needs a home. :)


jkeiser:
Realizing this probably goes better in the validateutf8_experiments repository. Will move.

jkeiser:
Closing to clean up PRs, but this is not lost.

PR #1049, opened 2020-07-18
Add equality operators to Object & Array iterators.

TkTech:
This adds operator== to the iterators, making them compatible with
pybind11's make_iterator().

Still missing lots of expected std::iterator functionality, but it's enough to get things working so I'm not maintaining a fork.


TkTech:
Closing in favor of #1050, which does more than "just make it work".

PR #1071, opened 2020-07-25
Add a quick example at the top of simdjson.h

pps83:


include/simdjson.h
```diff
@@ -5,6 +5,33 @@
  * @mainpage
  *
  * Check the [README.md](https://github.com/lemire/simdjson/blob/master/README.md#simdjson--parsing-gigabytes-of-json-per-second).
+
```
jkeiser, line 8: Having an example on the front page sounds great!

I think we want to keep the * in front of these lines, so that people can see what's code and what's comments. markdown  lets you put code in if you indent it (so add four spaces, like `  *     #include "simdjson.h"`

Also, period after "examples."

jkeiser:
Apologies for the delayed reviews @pps83 ... we've both been a little heavily loaded lately.

include/simdjson.h
pps83: I updated the comment, not sure if that's the way it should be done. Can you please show what you mean if last update isn't correct.

jkeiser: Yep! Thanks!

PR #1156, opened 2020-09-07
Remove unnamed namespaces from ondemand

jkeiser:
This makes it so you can use ondemand without having to enclose your own classes in an anonymous namespace. To do this I had to change the main unnamed namespace in src/ from <unnamed namespace>::SIMDJSON_IMPLEMENTATION to SIMDJSON_IMPLEMENTATION::<unnamed namespace>. This does not appear to have changed instruction counts or performance of the parse benchmark.

### DIFF
--- a/benchmark/kostya/iter.h
+++ b/benchmark/kostya/iter.h
@@ -6,7 +6,6 @@
 #include "kostya.h"
 
 namespace kostya {
-namespace {
 
 using namespace simdjson;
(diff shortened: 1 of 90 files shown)


jkeiser:
Not entirely on purpose, this (with some fixes for g++ 7) was already merged :)

jkeiser:
Except it wasn't :)

jkeiser:
Merging, though.

PR #1163, opened 2020-09-10
fuzz atpointer

pauldreik:
This adds a fuzzer for at_pointer() which recently had a bug.

The https://github.com/simdjson/simdjson/issues/1142 bug had been found with this fuzzer

Also, it polishes the github action job:
 - cross pollinate the fuzzer corpora (lets fuzzers reuse results from other fuzzers)
 - use github action syntax instead of bash checks
 - only run on push if on master

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -1,11 +1,9 @@
-name: Run fuzzers on stored corpus and test it with valgrind
+name: Fuzz and run valgrind
 
-# In the case of a pull request happening at the same time as a cron
-# job, there is a risk two jobs run at the same time. Therefore,
-# the corpus is only uploaded for the master branch. Pull requests will
-# fuzz for a short while, but the results are not uploaded.
 on:
   push:
+    branches:
+      - master
   pull_request:
   schedule:
     - cron: 23 */8 * * *
@@ -18,6 +16,7 @@ jobs:
       implementations: haswell westmere fallback
       UBSAN_OPTIONS: halt_on_error=1
       MAXLEN: -max_len=4000
+
     steps:
     - name: Install packages necessary for building
       run: |
@@ -28,24 +27,29 @@ jobs:
         sudo ./llvm.sh 9
 
     - uses: actions/checkout@v1
+
     - name: Create and prepare the initial seed corpus
       run: |
         fuzz/build_corpus.sh
         mv corpus.zip seed_corpus.zip
         mkdir seedcorpus
         unzip -q -d seedcorpus seed_corpus.zip
+
     - name: Download the corpus from the last run
       run: |
         wget --quiet https://dl.bintray.com/pauldreik/simdjson-fuzz-corpus/corpus/corpus.tar
         tar xf corpus.tar
         rm corpus.tar
+
     - name: List clang versions
       run: |
         ls /usr/bin/clang*
         which clang++
         clang++ --version
+
     - name: Build all the variants
       run: fuzz/build_fuzzer_variants.sh
+
     - name: Run the fast fuzzer (release build, default implementation, to explore fast)
       run: |
         set -eux
@@ -55,6 +59,7 @@ jobs:
           others=$(find out -type d -not -name $fuzzer -not -name out -not -name cmin)
           build-fast/fuzz/fuzz_$fuzzer out/$fuzzer $others seedcorpus -max_total_time=30 $MAXLEN
         done
+
     - name: Run the slow fuzzer (sanitizer+asserts, good at detecting errors)
       run: |
         set -eux
@@ -67,6 +72,7 @@ jobs:
           done
           echo now have $(ls out/$fuzzer |wc -l) files in corpus        
         done
+
     - name: Minimize the corpus with the fast fuzzer on the default implementation
       run: |
         set -eux
@@ -78,16 +84,19 @@ jobs:
           rm -rf out/$fuzzer
           mv out/cmin/$fuzzer out/$fuzzer
         done
+
     - name: Package the corpus into an artifact
       run: |
         for fuzzer in $allfuzzers; do      
           tar rf corpus.tar out/$fuzzer
         done
+
     - name: Save the corpus as a github artifact
       uses: actions/upload-artifact@v1
       with:
         name: corpus
         path: corpus.tar
+
     # This takes a subset of the minimized corpus and run it through valgrind. It is slow,
     # therefore take a "random" subset. The random selection is accomplished by sorting on filenames,
     # which are hashes of the content.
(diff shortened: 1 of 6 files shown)


pauldreik:
This is now ready for review and hopefully merging. @lemire, would you like to review?

jkeiser:
Apologies, I can't right now; work is distracting me from open source :)

I'll take a look tonight, though, if it hasn't been gotten to!

pauldreik:
> Apologies, I can't right now; work is distracting me from open source :)

Ha ha, will quote you on that! 

In the unlikely case you have no questions and think it can be merged as is, please go ahead!

.github/workflows/fuzzers.yml
```diff
@@ -28,82 +27,105 @@ jobs:
...
-    - name: Run the minimized corpus through valgrind (replay build, default implementation)
+
+    # This takes a subset of the minimized corpus and run it through valgrind. It is slow,
+    # therefore take a "random" subset. The random selection is accomplished by sorting on filenames,
+    # which are hashes of the content.
+    - name: Run some of the minimized corpus through valgrind (replay build, default implementation)
       run: |
         for fuzzer in $allfuzzers; do      
-          find out/$fuzzer -type f |sort|xargs -n40 valgrind build-replay/fuzz/fuzz_$fuzzer 2>&1|tee valgrind-$fuzzer.txt
+          find out/$fuzzer -type f |sort|head -n200|xargs -n40 valgrind build-replay/fuzz/fuzz_$fuzzer 2>&1|tee valgrind-$fuzzer.txt
```
jkeiser, line 106: How many items are in the corpus? Is there a way we can run all of them through valgrind, but in parallel, or are we going to increase planetary carbon emissions if we do that :)

.github/workflows/fuzzers.yml
```diff
@@ -28,82 +27,105 @@ jobs:
...
     - name: Build all the variants
       run: fuzz/build_fuzzer_variants.sh
+
     - name: Run the fast fuzzer (release build, default implementation, to explore fast)
       run: |
         set -eux
         for fuzzer in $allfuzzers; do
           mkdir -p out/$fuzzer # in case this is a new fuzzer, or corpus.tar is broken
-          build-fast/fuzz/fuzz_$fuzzer out/$fuzzer seedcorpus -max_total_time=30 $MAXLEN
+          # get input from everyone else (corpus cross pollination)
```
jkeiser, line 58: What's the reasoning here? Assume I have a 6-year-old's understanding of fuzzing where some superintelligent AI somewhere generated a bunch of JSON and a bunch of JSON pointers to go with them, and then it runs each of them one by one. How does corpus cross pollination work?

.github/workflows/fuzzers.yml
```diff
@@ -28,82 +27,105 @@ jobs:
...
-          echo uploading each artifact twice, otherwise it will not be published
-          curl -T corpus.tar -upauldreik:${{ secrets.bintrayApiKey }} https://api.bintray.com/content/pauldreik/simdjson-fuzz-corpus/corpus/0/corpus/corpus.tar";publish=1;override=1"
-          curl -T corpus.tar -upauldreik:${{ secrets.bintrayApiKey }} https://api.bintray.com/content/pauldreik/simdjson-fuzz-corpus/corpus/0/corpus/corpus.tar";publish=1;override=1"
-          curl -T valgrind.tar -upauldreik:${{ secrets.bintrayApiKey }} https://api.bintray.com/content/pauldreik/simdjson-fuzz-corpus/corpus/0/corpus/valgrind.tar";publish=1;override=1"
-          curl -T valgrind.tar -upauldreik:${{ secrets.bintrayApiKey }} https://api.bintray.com/content/pauldreik/simdjson-fuzz-corpus/corpus/0/corpus/valgrind.tar";publish=1;override=1"
-        else
-          echo "not on master, won't upload to bintray"
-        fi
+        echo uploading each artifact twice, otherwise it will not be published
+        curl -T corpus.tar -upauldreik:${{ secrets.bintrayApiKey }} https://api.bintray.com/content/pauldreik/simdjson-fuzz-corpus/corpus/0/corpus/corpus.tar";publish=1;override=1"
```
jkeiser, line 124: Not at all something we need to deal with now, but: should we / can we move the corpus somewhere owned collectively by simdjson, so that you don't have to bear the weight of the world forever and ever?

fuzz/FuzzUtils.h
```diff
@@ -0,0 +1,17 @@
+#ifndef SIMDJSON_FUZZUTILS_H
+#define SIMDJSON_FUZZUTILS_H
+
+#include <cstdint>
+
+// view data as a byte pointer
+template <typename T> inline const std::uint8_t* as_bytes(const T* data) {
+  return static_cast<const std::uint8_t*>(static_cast<const void*>(data));
```
jkeiser, line 8: Will `reinterpret_cast<const uint8_t*>` do the same thing? (Note: when I'm asking questions, it's usually because I don't know the answer (at least not for sure). I try not to *imply* a change, so if I don't say "we should [probably] change X." I'd rather not waste your time trying to guess what people mean when you could be coding!

fuzz/fuzz_atpointer.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // Inputs missing the separator line will get an empty json pointer
+    // but the all the input put in the document string. This means
+    // test data from other fuzzers that take json input works for this fuzzer
+    // as well.
+    const auto fd=split(as_chars(Data),Size);
+
+    simdjson::dom::parser parser;
+
+    // parse without exceptions, for speed
+    auto res=parser.parse(fd.json_doc.data(),fd.json_doc.size());
```
jkeiser, line 48: The idiomatic way to do this (in noexcept land) is to use error chaining:

```c++
error_code error = parser.parse(fd.json_doc.data(), fd.json_doc.size()).at_pointer(fd.json_pointer).get_string().error();
if (error) { return 0; }
```

It can be broken up if you want, but just FYI :) Error chaining lets you 

Also, most places we don't store the result and then get it--usually just error_code error = parse.X().get(y).  It's not necessary, but it'll make it easier for people to jump around the codebase. Plus, people will copy/paste/learn from the code anywhere, so making everything use the best idioms seems best :)

fuzz/fuzz_atpointer.cpp
```diff
@@ -0,0 +1,68 @@
...
+ * @return
+ */
+FuzzData split(const char *Data, size_t Size) {
+
+    using namespace std::literals;
+    constexpr auto sep="\n~~~\n"sv;
+
+    std::string_view all(Data,Size);
+    auto pos=all.find(sep);
+    if(pos==std::string_view::npos) {
```
jkeiser, line 26: Are there `SIMDJSON_PADDING` extra bytes after the end of the document? It looks like we just take whatever's being passed in, and I don't know whether the LLVM fuzzer pads the input ...

fuzz/fuzz_atpointer.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // test data from other fuzzers that take json input works for this fuzzer
+    // as well.
+    const auto fd=split(as_chars(Data),Size);
+
+    simdjson::dom::parser parser;
+
+    // parse without exceptions, for speed
+    auto res=parser.parse(fd.json_doc.data(),fd.json_doc.size());
+    if(res.error())
+        return 0;
```
jkeiser, line 50: What is a nonzero return value supposed to mean in a fuzzing scenario? In what cases can / should this function fail, and how do you denote that?

jkeiser (commented):
In general, it looks good. I'm still a level 1 fuzzing initiate, so I had a lot of questions that probably have answers :)

.github/workflows/fuzzers.yml
pauldreik: Roughly a thousand per fuzzer. You can download the corpus here from here: go to the [actions tab](https://github.com/simdjson/simdjson/actions), select the fuzzing job, artifacts, select corpus.
Yes one could run them in parallel but I do not think github action has more than one core so it would not help.
The valgrind run is less important when the undefined+address sanitizers are used since they detect many (if not all) errors valgrind would have found.

.github/workflows/fuzzers.yml
pauldreik: There is no magic here!
Most of the fuzzers use the same input (json files), but do different things with the parsed result. If one fuzzer has found interesting input, it will likely be useful to the other fuzzers. Corpus cross pollination is to try input from the other fuzzers and see if it benefits a particular fuzzer. This way, fuzzing effort is accumulated in case the input format is the same. When adding a new fuzzer to the mix, like this pull request does, it gets the benefit from the past effort of all the other fuzzers.

.github/workflows/fuzzers.yml
pauldreik: We could setup a simdjson account on bintray, but there is no rush. I do not use my bintray account for anything else, and it is good that I can upload a corpus manually if I want. For instance now, when I have spent a cpu month or so (compared to the 15 minutes of fuzzing per day made by the github action). One could also upload the corpus from oss fuzz which is not public, but accessible by us.

fuzz/fuzz_atpointer.cpp
pauldreik: Ah, nice! I agree with the idea that all code should be treated as if it is example code.
In this case, I am however interested in being able to inspect the intermediate result in a debugger. Plus it looks weird in non-fuzzing code asking for the string but only extract the error?

fuzz/fuzz_atpointer.cpp
pauldreik: nonzero return is reserved for future use. See here: https://llvm.org/docs/LibFuzzer.html#fuzz-target
Failure "on purpose" is indicated by abort(), assert() or throwing. Anything that interrupts the program. 
In the usual case, an exception means a normal error (such as invalid input) was detected and should be ignored by the fuzzer. Hence using try catch if using the throwing api of simdjson.

fuzz/fuzz_atpointer.cpp
pauldreik: There is no padding added by libFuzzer, which is a good feature. It means address sanitizer is likely to catch illegal reads. However, I see that I the parse() function mandates padded input! How could this not be detected when running the fuzzer? I will have to investigate this and come back.

pauldreik:
Thanks for the review, hope you got some answers! Keep asking.

I will have to find out why the lack of padding was not detected and come back.

pauldreik:
Mystery resolved, it is the default value of realloc_if_needed that fixes it. If I set it to false, address sanitizer catches the overread it in less than a second.

Speaking about example code that should be correct. Here is a mention of parse() which does not seem to respect the padding:

https://github.com/simdjson/simdjson/blob/33c97452012f7ba05812b33250281856f4f14de6/singleheader/simdjson.h#L2071-L2081

The doxygen document says padding is mandatory: 
https://github.com/simdjson/simdjson/blame/master/doc/basics_doxygen.md#L87-L88

...but the "mystery" is resolved further down in simdjson.h:

https://github.com/simdjson/simdjson/blob/7fc07e2d5ea57ceb0c8f99aa2ac205004d0d7ef3/singleheader/simdjson.h#L3402-L3408

I think this documentation string is confusing. The reason one does not have to think about padding is because the last parameter realloc_if_needed defaults to true.  I also believe that the text about the extra bytes being allocated is necessary but not have to be initialized is not true - reading from uninitialized data is undefined behaviour in C++.

May I suggest the following changes:

In [simdjson.h](https://github.com/simdjson/simdjson/blob/7fc07e2d5ea57ceb0c8f99aa2ac205004d0d7ef3/singleheader/simdjson.h#L3402-L3408) (I mean changing the underlying header before amalgamization):
```
### Buffer Padding
   *
   * If the realloc_if_needed is set to true (the default), the incoming buffer does not have to be
   * padded. If realloc_if_needed is false, the buffer has to be padded with at least 
   * SIMDJSON_PADDING bytes at the end.
   * Supplying a padded buffer is more efficient, because it potentially avoids an extra copy
   * which is beneficial for performance.
```
Perhaps there are other reasons for being more efficient - I guess the handling of reading data into simd registers is easier if one does not have to worry about the end?

In [doc/basics_doxygen.md](https://github.com/simdjson/simdjson/blame/master/doc/basics_doxygen.md#L87-L88):
```
Json can be parsed by reading from a string:
std::string str("[1,2,3]");
parser.parse(str);
If the input is padded with at least SIMDJSON_PADDING bytes, the parser can work more
efficiently:
parser.parse("[1,2,3]"_padded);
```
I can imagine that having the input coming from a large string or vector of char, the fastest way would be to resize the string or vector in place rather than letting a copy be made internally. But that would not fit under the basics section, perhaps?

jkeiser:
Yeah, we should have the document talk about what happens by default, and then specify what happens with realloc_if_needed=false as the "exception." Right now it's the other way around, and I can see why that would be confusing.

pauldreik:
Yep. But that's for another issue/pull request! Merging this now.

fuzz/FuzzUtils.h
pauldreik: Sorry, missed to answer this! Yes, reinterpret_cast casts anything to anything (like a c-style cast). Therefore one tries to avoid it since it is so powerful. Also, i think the function name conveys meaning and is easier to read.

PR #1179, opened 2020-09-18
Allow direct document iteration in ondemand

jkeiser:
This allows you to do `for (auto val : doc)` instead of `for (auto elem : doc.get_array())`. It's a small thing, but iterating over top-level arrays is common enough that I didn't want people to mis-learn how to use the API (you can say `for (auto elem : val)` anywhere else _except_ document).

The reason it was hard is that document *owns* the iterator, while value and array *reference* the iterator. The array_iterator marks the iteration as finished by releasing the iterator, but you can't do that with an owned iterator. This fixes the problem by having array_iterator<value>, array_iterator<array>, and array_iterator<document>, which call `finished_iterating()` on the value/array/document when it's finished. document just stores that state differently.

This is the only API design wart I know of, though there are still some bugs and safety issues I have plans to address.

### DIFF
--- a/include/simdjson/generic/ondemand/array-inl.h
+++ b/include/simdjson/generic/ondemand/array-inl.h
@@ -64,10 +64,26 @@ simdjson_really_inline array array::started(json_iterator_ref &&iter) noexcept {
   return array(std::forward<json_iterator_ref>(iter));
 }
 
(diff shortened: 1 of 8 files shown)


jkeiser:
(I'll merge this in short order, but want to keep the discipline of making PRs for each change since #947 is getting pretty huge.

PR #1189, opened 2020-09-26
add CI fuzzing on arm 64 bit

pauldreik:
This adds fuzzing on drone.io arm64

For some reason, leak detection had to be disabled. If it is enabled, the fuzzer falsely reports a crash at the end of fuzzing.

Closes: #1188


lemire (approved):
This sure looks good.

PR #1235, opened 2020-10-15
demonstrate parse_unsigned is unused

pauldreik:
It seems like parse_unsigned is not used by anything running in CI.

This PR is here to prove that that either that function is not used and should be removed, or something needs to be added to CI.


jkeiser:
Yes, parse_large_unsigned is presently unused. I'm fine with removing it and resurrecting it later; it is designed to handle integers we know are going to be large, like in the case of Twitter IDs.

Issue #74, opened 2019-02-25
Port to POWER9


lemire:
There is demand to port this library to POWER9 

https://github.com/TkTech/pysimdjson/issues/9

I own a POWER9 box that I can make remotely available to someone interested in helping with the port.

cc @olgabot @TkTech

TkTech:
Definitely happy to support it on the python side of things if a port is available. Need to look around for anyone hosting a buildbot on ppc for the simdjson and pysimdjson builds & tests.

danlark1:
Hey, I am gonna outsource it to one of the students at the faculty of computer science in HSE university https://www.hse.ru/en/ under my guidance as a university project, I have enough experience with dealing SIMD in PPC, for example, the transition layer for x86_64 -> PPC [intrinsics](https://github.com/catboost/catboost/blob/master/library/cpp/sse/powerpc.h) so I believe I can manage this.

lemire:
@danlark1 Please get in touch. I think that @jkeiser and myself will be happy to help.

I have a POWER8 server, if it helps.

lemire:
(I meant POWER9 in my comment above.)

danlark1:
@user959 is going to take this issue and start working shortly

lemire:
Great.

danlark1:
@TkTech we merged the ppc64 support, in order to build specifically for power8 and power9 at the same time you can use `"-DSIMDJSON_IMPLEMENTATION=ppc64;fallback"`, if you want to support only power9 I suggest adding flags `-DCMAKE_CXX_FLAGS="-maltivec -mcpu=power9 -mtune=power9" -DCMAKE_C_FLAGS="-maltivec -mcpu=power9 -mtune=power9"`, the implementation is the same, the compiler optimizations are a bit different. The performance difference is around 10%

lemire:
Will be in 0.7.

@danlark1 Your name was added as a contributor.

PR #1254, opened 2020-10-26
PPC64 support

danlark1:
Adding PPC64/Altivec support for SimdJSON, it should work for POWER8 and POWER9 at the same time and I tested both architectures with both endianness types, it should fully cover #74

# Summary

I supported both big-endian and little-endian compilation modes -- whenever the library is compiled, it would be correctly compiled according to the architecture. Windows supports only very old 32 bit modes, for now I decided not to support it and waste time (however, in bitmanipulation.h I remained the macros for consistency). Small summary can also be found [here](https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/PowerPC-AltiVec_002fVSX-Built-in-Functions.html).

# Overview

SIMD set for PPC64 is called [AltiVec](https://en.wikipedia.org/wiki/AltiVec) and it defines the types with `__vector` as a prefix. For example, `__vector signed char`, `__vector unsigned char`, `__vector int`, etc. They are all 16 bytes. The access of the elements is just a bracket operator. SIMD instructions start with `vec_*` and the reference you can find [here](https://www.ibm.com/support/knowledgecenter/SSXVZZ_13.1.3/com.ibm.xlcpp1313.lelinux.doc/compiler_ref/vec_intrin_cpp.html). The operations are pretty similar for arithmetic (including saturation) and comparisons, you should be able to understand while reading them. I also defined `__m128i` as `__vector unsigned char` because it is easier to see the similarities between the westmere and altivec implementations, however, I am free to remove it.

# Differences with SSE for SIMDJSON

AltiVec does not have pure [_mm_alignr_epi8](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_alignr_epi8) for `prev` predicate, however it can be replicated with [`vec_sld(a, b, c)`](https://www.ibm.com/support/knowledgecenter/SSXVZZ_13.1.3/com.ibm.xlcpp1313.lelinux.doc/compiler_ref/vec_sld.html) which shifts by `c` two concatenated vectors `a` and `b`. Given the fact that `prev` predicate has a template parameter which is less than 16, for big-endian it is just `vec_sld(prev_chunk, chunk, 16 - N)`, for little-endian we need to swap bytes with [`vec_reve`](https://www.ibm.com/support/knowledgecenter/SSXVZZ_13.1.3/com.ibm.xlcpp1313.lelinux.doc/compiler_ref/vec_reve.html) for `prev_chunk`, `chunk` and `result`. I haven't come up with anything better for this.

AltiVec also does not have pure [_mm_shuffle_epi8](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_shuffle_epi8), however, it has [`vec_perm(a, b, c)`](https://www.ibm.com/support/knowledgecenter/SSGH3R_16.1.0/com.ibm.xlcpp161.aix.doc/compiler_ref/vec_perm.html) which chooses by least 5 significant bits of each byte of vector `c` which byte choose from 32 of concatenation from `a` and `b` and `vec_perm(a, a, mask)` should almost work except when the bytes are negative. Fortunately, `lookup_16` never uses negative bytes and `vec_perm` is working interchangeably. In `compress` there is a table `pshufb_combine_table` with negative bytes, I decided to patch a bit this table by replacing `0x80` to `0xff` and using `vec_perm(a, splat(0), c)` as it will choose zero bytes. This will work for all implementations, SSE, NEON and AltiVec at the same time.

Also there were problems with [`_mm_movemask_epi8`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_movemask_epi8), I replaced it with [`vec_vbpermq(a, b)`] which by index mask `b` takes all these bits from `a`. In the end, it began to be the following:

```cpp
  simdjson_really_inline int to_bitmask() const {
    __vector unsigned long long result;
    // Bits number 120, 112, 104, 98, down to 0
    const __m128i perm_mask = {0x78, 0x70, 0x68, 0x60, 0x58, 0x50, 0x48, 0x40,
                               0x38, 0x30, 0x28, 0x20, 0x18, 0x10, 0x08, 0x00};

    result = ((__vector unsigned long long)vec_vbpermq((__m128i)this->value,
                                                       (__m128i)perm_mask));
#ifdef __LITTLE_ENDIAN__
    return static_cast<int>(result[1]);
#else
    return static_cast<int>(result[0]);
#endif
  }
```

## Nonobvious details

For `prefix_xor` function I decided to use the fallback version, however there is a similar to pclmul instruction [`vec_pmsum_be`](https://www.ibm.com/support/knowledgecenter/SSXVZZ_13.1.3/com.ibm.xlcpp1313.lelinux.doc/compiler_ref/vec_pmsum_be.html), however, it is not defined for polynomial multiplication for unsigned long long types. gcc returns `__vector __int128`, clang returns `__vector unsigned long long` (which, obviously, has 2 elements), the code is strange but the following

```cpp
simdjson_really_inline uint64_t prefix_xor(uint64_t bitmask) {
  __vector unsigned long long all_ones = {~0ull, ~0ull};
  __vector unsigned long long mask = {bitmask, 0};
  // Clang and GCC return different values for pmsum for ull so a bit strange
  // dispatch. Generally it is not specified by ALTIVEC ISA what is returned by
  // vec_pmsum_be.
#if defined(__LITTLE_ENDIAN__)
  return (uint64_t)(((__vector unsigned long long)vec_pmsum_be(all_ones, mask))[0]);
#else
  return (uint64_t)(((__vector unsigned long long)vec_pmsum_be(all_ones, mask))[1]);
#endif
}
```

llvm-mca showed that it provides a better codegen than the fallback implementation of that. Some versions of gcc, for example, on my test server and in QEMU [miscompiled that thing](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91275). Performance is almost the same, I decided to proceed with a safe version.

One more strange thing is the undef of `vector` and `bool`, this comes from [`altivec.h`](https://github.com/gcc-mirror/gcc/blob/793b38da83ed2bb876447634fcdd705ce1266f28/gcc/config/rs6000/altivec.h#L42) header, for compatibility I decided to undef these things, though code looks a bit strange. 

# Performance

Benchmarks. Only recent yyson is faster for Power8, for Power9 I'd say this is a tie, AltiVec simdjson is around 2.5 times faster than its fallback version

**POWER8:**
2020-10-26-232141_1451x865_scrot

**POWER9:**
gbps-1

# CI

2020-10-26-225217_1532x568_scrot

Travis CI is green https://travis-ci.org/github/danlark1/simdjson/builds/739108577 please add it to the consequent pull requests here as I don't see it

# Comments

Sorry for the trailing whitespaces to be removed, my text editor does it automatically on save, I noticed that very late and already added a bunch of documentation, if you are very uncomfortable with them, I'll remove.

Also, AltiVec is very picky on the casts and type safety, you will see a lot of C cast styles, that is intentional and needed to silence all the compilers.

# How to test manually

I tested this with clang 10.0.1 and gcc 9.1.0.

Apart from bare metal in order to test it on a local machine you should obtain somehow the toolchain, for example for clang and gcc from apt or pacman ([simde](https://github.com/simd-everywhere/simde/blob/66d298f90f5934a881867969ff1210c7cf898770/test/docker/Dockerfile.ppc64el) does the same way):

After downloading a toolchain in order to cross-compile I used for big-endian clang:

```console
$ CC=clang CXX=clang++ cmake -DCMAKE_CXX_FLAGS="--target=ppc64-pc-linux-gnu --sysroot /usr/powerpc64-linux-gnu -fuse-ld=/usr/bin/powerpc64-linux-gnu-ld -Wno-unused-command-line-argument -maltivec -mvsx -mpower8-vector" -DCMAKE_C_FLAGS="--target=ppc64-pc-linux-gnu --sysroot /usr/powerpc64-linux-gnu -fuse-ld=/usr/bin/powerpc64-linux-gnu-ld  -Wno-unused-command-line-argument -maltivec -mvsx -mpower8-vector" -DCMAKE_SYSTEM_NAME=ppc64 -DSIMDJSON_IMPLEMENTATION=ppc64 ..
```

For little-endian

```console
$ CC=clang CXX=clang++ cmake -DCMAKE_CXX_FLAGS="--target=ppc64le-pc-linux-gnu --sysroot /usr/powerpc64le-linux-gnu -fuse-ld=/usr/bin/powerpc64le-linux-gnu-ld -Wno-unused-command-line-argument" -DCMAKE_C_FLAGS="--target=ppc64le-pc-linux-gnu --sysroot /usr/powerpc64le-linux-gnu -fuse-ld=/usr/bin/powerpc64le-linux-gnu-ld -Wno-unused-command-line-argument" -DCMAKE_SYSTEM_NAME=ppc64 -DSIMDJSON_IMPLEMENTATION=ppc64 ..
```

With gcc it was a bit easier

Big-endian:

```console
$ CC=powerpc64-linux-gnu-gcc CXX=powerpc64-linux-gnu-g++ cmake -DCMAKE_CXX_FLAGS="-maltivec -mvsx -mpower8-vector" -DCMAKE_C_FLAGS="-maltivec -mvsx -mpower8-vector" -DCMAKE_SYSTEM_NAME=ppc64 -DSIMDJSON_IMPLEMENTATION=ppc64 ..
```

Little-endian:

```console
$ CC=powerpc64le-linux-gnu-gcc CXX=powerpc64le-linux-gnu-g++ cmake -DCMAKE_SYSTEM_NAME=ppc64 -DSIMDJSON_IMPLEMENTATION=ppc64 ..
```

If you want to test it for power9 or power8 specifically, use `-march=power8` and `-march=power9` options in clang and for gcc `-mcpu=power8` and `-mcpu=power9`. Add them in CMAKE_CXX_FLAGS and CMAKE_C_FLAGS.

Then I used qemu for testing

```console
$ QEMU_LD_PREFIX=/usr/powerpc64-linux-gnu qemu-ppc64 basictests # big-endian
$ QEMU_LD_PREFIX=/usr/powerpc64le-linux-gnu qemu-ppc64le basictests # little-endian
```

The third argument can be any test that is compiled in the library.


lemire:
@danlark1 So you do not have a POWER 9 processor? Please email me at lemire@gmail.com and I will see about giving you remote access.


1. Can you check that QEMU actually validates your kernel/implementation and not, say, the fallback implementation? 
2. Can we use QEMU in CI?

It is not going to be possible to support a platform if we cannot test for it. That is, if we cannot test it, we cannot document it as a supported platform.


> Sorry for the trailing whitespaces to be removed, my text editor does it automatically on save, I noticed that very late and already added a bunch of documentation, if you are very uncomfortable with them, I'll remove.

That's not a big deal in this instance. Note that it is possible to commit only the non-trivial changes.

https://stackoverflow.com/questions/13793241/make-git-undo-any-whitespace-only-changes/45486981

What I typically do before I commit, is that I do `git diff` and if I see a file that should not have changed, I do `git checkout thefile` and I undo the changes.

lemire:
@gerrith3 @srinivas32 Any chance you could help?

lemire:
@danlark1 I think travis supports POWER: https://github.com/RoaringBitmap/CRoaring/blob/master/.travis.yml#L3

danlark1:
> 1. Can you check that QEMU actually validates your kernel/implementation and not, say, the fallback implementation?

Yeah, I am pretty sure it validates the kernel implementation as I was debugging bugs for quite some time

> 2. Can we use QEMU in CI?

We definitely can, also I found that only travis ci supports ppc64le hosts. Unfortunately none of oCIs support big endian architectures. What would you prefer? Going for qemu or using travis for this? I see a bunch of CIs used and going for one more might not be an obvious decision

> What I typically do before I commit, is that I do `git diff` and if I see a file that should not have changed, I do `git checkout thefile` and I undo the changes.

There were just too many changes, I gave up a bit, thanks for link, btw :)

lemire:
>  Going for qemu or using travis for this? I see a bunch of CIs used and going for one more might not be an obvious decision

If we could run a QEMU check, testing at least just basictests, that would be pretty good. If you could propose something, that would be great.

> Unfortunately none of oCIs support big endian architectures. 

For better or for worse, LE has won, at least in the linux world. I don't think I have any BE system.

danlark1:
> If we could run a QEMU check, testing at least just basictests, that would be pretty good. If you could propose something, that would be great.

Okay, I will try to add qemu for all tests. Travis looks inviting, though, not sure what will be easier.

lemire:
@danlark1 If travis works, then it is definitively better.

gerrith3:
@lemire @danlark1 Yes - Travis CI does now support Power hardware, specifically ppc64le only - not big endian which IBM has abandoned from a Linux perspective.  The hardware can support both big/little but all of the Travis CI back end is Power9 hardware with Linux, running LXD containers, so you should be able to get valid testing of Altivec instructions.

lemire:
@gerrith3 Very useful, thank you.

danlark1:
Unfortunately, in travis there is old cmake for this and all workarounds suggested [here](https://travis-ci.community/t/outdated-versions-of-cmake/8331) don't work properly. Facebook with rocksdb faced the same issue https://github.com/facebook/rocksdb/pull/7381

Will see what I can do here

lemire:
@danlark1 Solving CI issues is my favorite hobby. :-)

danlark1:
> @danlark1 Solving CI issues is my favorite hobby. :-)

Yeah, I thought it would be easier but I believe now the config is in good shape unless gcc starts ooming again for singleheader.

Also gcc7 is falling into ICE so PPC64 will be supported for clang6+ and gcc8+, I believe that's ok

gerrith3:
@danlark1 I think for ppc64le, gcc7 is really, really old.  So excluding it on Power is probably a good idea.

danlark1:
2020-10-26-225217_1532x568_scrot

CI is green https://travis-ci.org/github/danlark1/simdjson/builds/739108577 please add it to the consequent pull requests here as I don't see it

Benchmarks. Only recent yyson is faster, around 2.5 times faster than the fallback version 

2020-10-26-232141_1451x865_scrot

danlark1:
> @danlark1 I think for ppc64le, gcc7 is really, really old. So excluding it on Power is probably a good idea.

@gerrith3 I sometimes see problems in miscompilation for gcc in vec_pmsum_be, were there any issues you can remember? I am falling back to a safe version (and it turned out the performance is almost not noticeable)

jkeiser:
Thanks for the benchmarks! Note that yyjson being faster shouldn't stop us from merging. If we end up using yyjson as a fallback, we might revisit whether to keep the implementation on., but even then, having this implementation around gives us a base to iterate on.

It does underscore that we need to look _really_ close at what yyjson is doing. Something is going on there :)

gerrith3:
> > @danlark1 I think for ppc64le, gcc7 is really, really old. So excluding it on Power is probably a good idea.
> 
> @gerrith3 I sometimes see problems in miscompilation for gcc in vec_pmsum_be, were there any issues you can remember? I am falling back to a safe version (and it turned out the performance is almost not noticeable)

I'm going to ask my friend and colleague Paul Clarke @pacman to see if he remembers anything...

danlark1:
> > > @danlark1 I think for ppc64le, gcc7 is really, really old. So excluding it on Power is probably a good idea.
> > 
> > 
> > @gerrith3 I sometimes see problems in miscompilation for gcc in vec_pmsum_be, were there any issues you can remember? I am falling back to a safe version (and it turned out the performance is almost not noticeable)
> 
> I'm going to ask my friend and colleague Paul Clarke @pacman to see if he remembers anything...

Smallest example @pacman @gerrith3 

```
#include <altivec.h>
#include <stdint.h>
#include <bitset>
#include <iostream>
#include <cstring>
#include <random>
#include <byteswap.h>

uint64_t prefix_xor(uint64_t bitmask) {
  bitmask ^= bitmask << 1;
  bitmask ^= bitmask << 2;
  bitmask ^= bitmask << 4;
  bitmask ^= bitmask << 8;
  bitmask ^= bitmask << 16;
  bitmask ^= bitmask << 32;
  return bitmask;
}

uint64_t prefix_xor_altivec(uint64_t bitmask) {
    __vector unsigned long long all_ones = {~0ull, ~0ull};
    __vector unsigned long long mask = {bitmask, 0};
#if defined(__LITTLE_ENDIAN__)
  return (uint64_t)(((__vector unsigned long long)vec_pmsum_be(all_ones, mask))[0]);
#else
  return (uint64_t)(((__vector unsigned long long)vec_pmsum_be(all_ones, mask))[1]);
#endif
}

int main() {
    std::mt19937_64 e2(123);
    std::uniform_int_distribution<uint64_t> dist(0, std::numeric_limits<uint64_t>::max());
    size_t j = 0;
    for (size_t i = 0; i < 200; ++i) {
        auto random_num = dist(e2);
        auto a = prefix_xor(random_num);
        auto b = prefix_xor_altivec(random_num);
        if (a != b) {
          std::cout << std::bitset<64>(a) << "\n" << std::bitset<64>(b) << std::endl;;
          std::cout << (a == b) << ' ' << i << ' ' << j++ << ' ' << random_num << std::endl;
        }
    }
}
```

```
g++ --version
g++ (Debian 8.3.0-6) 8.3.0
g++ A.cpp -O3 -o A
./A
...
```
"print 94 differences (out of 200)". Last lines:

```
1010001011110000011000110100000010101101111101101110011000110001
0101110100001111100111001011111101010010000010010001100111001110
```
The bits are exactly complementary. In CI with gcc 8.4 and 9.3 all is good. With clang also all is good.

```
clang++ --version
clang version 7.0.1-8 (tags/RELEASE_701/final)
```

danlark1:
@lemire I believe all is ready, please review :+1: so that I can iterate

danlark1:
@gerrith3, yeah, this is an issue https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91275, found it!

danlark1:
@jkeiser 

Haha, that was my fault also, the previous benchmarks were for power8, for power9 we have a much better situation

gbps-1

lemire:
@danlark1 I am personally happy to merge this. We'll have to add your name to the contributors, but this can be done in a second step.

Let us wait till the tests turn green.

lemire:
Could you push something to see if travis will wake up?

danlark1:
> Could you push something to see if travis will wake up?

Done, seems that it started

lemire:
@danlark1 Yep. Good.

danlark1:
> @danlark1 Yep. Good.

@lemire can you cancel some useless jobs in appveyor so that it can proceed? My bad to include so many commits to fix travis

lemire:
I have cancelled a bunch, but I don't think that's your fault... I think something is going on at appveyor. It seems that the jobs got stuck. 

I'd be ok with merging this PR without appveyor's approval. Whatever harm you may have done to Visual Studio is surely fixable.

danlark1:
> I have cancelled a bunch, but I don't think that's your fault... I think something is going on at appveyor. It seems that the jobs got stuck.
> 
> I'd be ok with merging this PR without appveyor's approval. Whatever harm you may have done to Visual Studio is surely fixable.

Okay, I believe as a first iteration it is more than enough, we got pretty good results already reaching 1GB/s on many cases and beating sajson fully and yyjson in ~half cases.

ThinkOpenly:
> > > @gerrith3 I sometimes see problems in miscompilation for gcc in vec_pmsum_be, were there any issues you can remember? I am falling back to a safe version (and it turned out the performance is almost not noticeable)
> > 
> > I'm going to ask my friend and colleague Paul Clarke @ThinkOpenly to see if he remembers anything...
> 
> Smallest example @ThinkOpenly @gerrith3

> "prints 94 differences (out of 200)". Last lines:
> 
> ```
> 1010001011110000011000110100000010101101111101101110011000110001
> 0101110100001111100111001011111101010010000010010001100111001110
> ```
> 
> The bits are exactly complementary for all differences. In CI with gcc 8.4 and 9.3 all is good, with GCC 8.3.0 and 9.2.1 fail. With clang from version 6 to 10 also all is good. With `-O0` also all is good everywhere, this is (was?) definitely a miscompilation

I can reproduce the failure with
g++ (GCC) 8.2.1 20180813 (Advance-Toolchain-at12.0) [revision 263510]

But it works with:
g++ (GCC) 9.3.1 20191023 (Advance-Toolchain 13.0-2) [200e83ac36b7]
g++ (GCC) 10.2.1 20200818 (Advance-Toolchain 14.0-0) [30ffe7eef838]

I expect the next version of Advance Toolchain 12 will have GCC 8.4.

I should also note that GCC for POWER is incorporating support for compatibility with many of the Intel vector intrinsics (currently through SSSE3 and some SSE4.1). Not guaranteed to be optimal, but might be a reference for implementing some of the ones that are not 1-to-1 matches with the Power Vector Intrinsics, like _mm_alignr_epi8, _mm_shuffle_epi8, and_mm_movemask_epi8. There are open implementations for each of these in their respective files.

danlark1:
> > > > @gerrith3 I sometimes see problems in miscompilation for gcc in vec_pmsum_be, were there any issues you can remember? I am falling back to a safe version (and it turned out the performance is almost not noticeable)
> > > 
> > > 
> > > I'm going to ask my friend and colleague Paul Clarke @ThinkOpenly to see if he remembers anything...
> > 
> > 
> > Smallest example @ThinkOpenly @gerrith3
> 
> > "prints 94 differences (out of 200)". Last lines:
> > ```
> > 1010001011110000011000110100000010101101111101101110011000110001
> > 0101110100001111100111001011111101010010000010010001100111001110
> > ```
> > 
> > 
> > The bits are exactly complementary for all differences. In CI with gcc 8.4 and 9.3 all is good, with GCC 8.3.0 and 9.2.1 fail. With clang from version 6 to 10 also all is good. With `-O0` also all is good everywhere, this is (was?) definitely a miscompilation
> 
> I can reproduce the failure with
> g++ (GCC) 8.2.1 20180813 (Advance-Toolchain-at12.0) [revision 263510]
> 
> But it works with:
> g++ (GCC) 9.3.1 20191023 (Advance-Toolchain 13.0-2) [200e83ac36b7]
> g++ (GCC) 10.2.1 20200818 (Advance-Toolchain 14.0-0) [30ffe7eef838]
> 
> I expect the next version of Advance Toolchain 12 will have GCC 8.4.
> 
> I should also note that GCC for POWER is incorporating support for compatibility with many of the Intel vector intrinsics (currently through SSSE3 and some SSE4.1). Not guaranteed to be optimal, but might be a reference for implementing some of the ones that are not 1-to-1 matches with the Power Vector Intrinsics, like _mm_alignr_epi8, _mm_shuffle_epi8, and_mm_movemask_epi8. There are open implementations for each of these in their respective files.

I already found an issue with this https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91275 so we are good

Yeah, I know about that compatibility layer, I consulted it and fortunately I found many ways to optimize the things natively better

danlark1:
@lemire CI is green, feel free to merge

lemire:
Merging. I will be adding your name to the contributors.

PR #1267, opened 2020-10-31
fuzz with the intended clang version

pauldreik:
This builds the CI fuzzers with the intended clang version. It also allows users to set the clang version locally,
in case they need to.

It also switches the CI fuzzers to use an optimized sanitizer build, to do something oss-fuzz doesn't and get more done in the short time the CI fuzzer runs.


pauldreik:
Merging, because it only affects fuzzing and appveyor seems to be broken.

Issue #1264, opened 2020-10-29
enable/disable deprecated api from cmake


pauldreik:
There are some parts of the simdjson api that are deprecated, an example is [element::at(std::string_view)](https://github.com/simdjson/simdjson/blob/ed21875083baf858dd4b16a253e21755074df122/include/simdjson/dom/element-inl.h#L119).

It is good to let users know which parts they should not use and why - the deprecated attribute, as seen in the above example, works for that.
But it would be good to prevent it from being active in the source at all, by hiding it inside ifdefs. The benefits are:
- fuzzing coverage will ignore it, making a more accurate view of how well the library is fuzzed (simplifies where to guide fuzzer development effort).
- IDE-s will ignore it so it is not suggested when using completion
- the IDE will grey it out by default, reducing the mental burden of browsing the code

I suggest:
- cmake boolean config SIMDJSON_ENABLE_DEPRECATED_API which is "Off" by default
- when the above is enabled, sets macro SIMDJSON_ENABLE_DEPRECATED_API=1, otherwise nothing
- deprecated apis use #ifdef SIMDJSON_ENABLE_DEPRECATED_API

I volunteer for this, collecting feedback here before I start.

PR #1271, opened 2020-10-31
optionally disable deprecated apis

pauldreik:
This is a shot at fixing #1264.

I think that the deprecated apis should be off by default, but for now I set them to be on by default.

The fuzzers disable the deprecated apis.

I have tested that the amalgamation still works with/without the deprecated api.


include/simdjson/dom/parsedjson_iterator.h
```diff
@@ -266,5 +267,6 @@ class [[deprecated("Use the new DOM navigation API instead (see doc/basics.md)")
 };
 
 } // namespace simdjson
+#endif //#if SIMDJSON_ENABLE_DEPRECATED_API
```
jkeiser, line 270: Remove the #if, I think--generally we don't include that in the #endif matching comment.

jkeiser:
Because users use the amalgamated headers raw, without cmake, we need it to default to on, with something like

```c++
#ifndef SIMDJSON_ENABLE_DEPRECATED_API
#define SIMDJSON_ENABLE_DEPRECATED_API 1
#endif
```

That, or call it SIMDJSON_DISABLE_DEPRECATED_API :)

tests/basictests.cpp
```diff
@@ -522,12 +525,16 @@ namespace parse_api_tests {
     return parser_moving_parser() &&
            parser_parse() &&
            parser_parse_many() &&
+#ifdef SIMDJSON_ENABLE_DEPRECATED_API
```
jkeiser, line 528: Goodness, we *deeply* need to use a real test framework :)

jkeiser (changes requested):
With the minor change of making sure it defaults to on without cmake, I'm super happy with this! It makes a ton of sense. Thanks :)

include/simdjson/dom/parsedjson_iterator.h
pauldreik: sure thing!

tests/basictests.cpp
pauldreik: Yes....

pauldreik:
> With the minor change of making sure it defaults to on without cmake, I'm super happy with this! It makes a ton of sense. Thanks :)

Thanks for the review, I fix these things and merge.

tests/basictests.cpp
friendlyanon: Just a suggestion, but [doctest](https://github.com/onqtam/doctest) is pretty good!

PR #1272, opened 2020-10-31
fuzz more of generic/numberparsing.h

pauldreik:
include/simdjson/generic/numberparsing.h have several functions which are not reached. The ondemand fuzzer was extended to call those apis as well.


pauldreik:
Merging, this should not harm anyone.

PR #1280, opened 2020-11-02
Do not enable -fPIC when compiling on Windows

jeremyong:
Regardless of the compiler, Windows targets do not support -fPIC,
as position independent code is already implicitly enabled. Compiling
simdjson with Clang on Windows will error because -fPIC is an
unsupported option for target 'x86_64-pc-windows-msvc'.


jeremyong:
Incidentally, it may be worth doing a pass to see if other aspects of the build assume `NOT MSVC` is equivalent to `NOT WIN32`. Clang compilation on Windows is a fairly common workflow as well.

jeremyong:
The travis check that failed is due to a spurious github.com outage

image

lemire:
Merging.

We had disabled travis because it was always falling on its head. We re-enabled it for its support of POWER processors... but it is still poor as far as reliability goes.

xkszltl:
Should it be `POSITION_INDEPENDENT_CODE` instead of the hardcoded `-fPIC`?
- https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html#prop_tgt:POSITION_INDEPENDENT_CODE

xkszltl:
Could you help verifying this?
- https://github.com/simdjson/simdjson/pull/2383

Issue #1215, opened 2020-10-07
Support out-of-order key lookup in On Demand


jkeiser:
Presently, On Demand requires that fields in JSON are strictly ordered in the order you get them in the code: if you write `obj["x"] + obj["y"] + obj["z"]`, the JSON `{ "z": 3, "y": 2, "x": 1 }` will fail to parse--it will skip z and y while looking for x.

While that will always be the fast path, we should allow for JSON that is not in the same order.

We could have the user to tell us which kind of indexing they want when they get the object--i.e. get_ordered_object() or get_unordered_object(). Then we would know to either make a lookup table or save a mark at the beginning of the object.

jkeiser:
Fixed by #1351.

PR #1351, opened 2020-12-20
Make `object["field"]` order-insensitive in On Demand

jkeiser:
This makes field lookup order-insensitive by default in On Demand: out-of-order lookups will succeed.

This means this JSON can be properly processed with code that does `object["x"].get_double() + object["y"].get_double()`:

```json
[
  { "x": 1, "y": 2 },
  { "y": 2, "x": 1 }
]
```

The previous order-sensitive behavior can still be accessed with `object.find_field("field")`.

## Design

* `object.find_field("field")` does an order-sensitive search.
* `object.find_field_unordered("field")` does an order-insensitive search.
* `object["field"]` uses find_field_unordered().
* When fields are in order, it behaves exactly the same as find_field(), starting after "x" when it's looking for "y".
* When fields are out of order, the find_field() algorithm will reach the end without finding anything. At this point, find_field_unordered() cycles back to the beginning of the object and searches from there, stopping when it reaches the original (in order) starting point. If the field is still not found, it returns NO_SUCH_FIELD. This is what allows it to find "y" even though it's not after "x".

## Performance

This may cause minor performance regressions relative to order-sensitive lookup (there is certainly no reason to believe it will improve performance). There is now an OnDemandUnordered version of the LargeRandom benchmark, which is roughly 2% worse than the ordered version, a real difference caused almost entirely by increased instruction count:

| Benchmark | Throughput | Instructions | Branch Misses |
|---|---|---|---|
| LargeRandom&lt;OnDemand&gt; | 634.066 MB/s | 614,119,494 | 972,480 |
| LargeRandom&lt;OnDemandUnordered&gt; | 650.795 MB/s | 636,119,541 | 956,152 |

All benchmarks were updated to use `find_field()` since they are intended to be order-sensitive. This also let me verify that there is, as expected, no performance difference between the previous `[]` and the new `find_field()`.


jkeiser:
Fixes #1215.

jkeiser:
Had to disable a strict-overflow warning on GCC 7, where it warns when the line `if (depth() <= parent_depth) { return SUCCESS; }` gets optimized away. This can happen since parent_depth and depth() are intimately related and parent_depth is designed to be constant at any given point in the program.

jkeiser:
The remaining error is interesting. It's taking too long to compile basic tests with sanitize on.

I'll figure that one out, but I don't expect the code itself to change ... probably just break up the basic tests.

jkeiser:
Oh! The reason make_pair(obj["w"], obj["h"]) fails is that it creates a `pair<value, value>`, which is then implicitly converted to `pair<uint64_t, uint64_t>`. This means the cursor is at "h" by the time we try to parse w, which is obviously a problem. A debug build would catch this in a more intelligible way, with a runtime assert.

I'd like to figure out how to make release builds do this; I think it can be done without performance impact. I'll try in a separate checkin, this bug existed before this PR.

jkeiser:
OK, I broke up ondemand_basictests so that each test namespace is in its own executable, to reduce individual compile speeds and allow sanitize builds to complete.

jkeiser:
mingw failed now because it referenced ondemand_basictests explicitly, which no longer exists. I've added automatic targets `all_tests`, `acceptance_tests`, etc. and also added a `no_mingw` test label that can be used to exclude tests we don't want to run on mingw.

jkeiser:
Actually, first let's just see if all the acceptance tests *pass* on mingw. Then we'll just run 'em.

jkeiser:
@lemire all the acceptance tests pass on mingw, it seems (happy day!). You were [explicitly only running a subset of tests](https://github.com/simdjson/simdjson/pull/1004/files#diff-3f685e4f868e9c02b6356591261b0aabd5f2508201cf7087ffdc6bc2fbeddd15R49-R54) to keep it fast ... but it is 3-7 minutes right now. Is that fast enough to enable them all?

jkeiser:
(Note: the fact that it's fast now might be because I added an `acceptance_tests` target to just build those tests.)

jkeiser:
Well :) Adding a few more tests put the tests over the size threshold and required a little more work than otherwise necessary, but the world is better for it :)

lemire:
@jkeiser What is your take on these Win32 errors?

```
The following tests FAILED:
4427	 26 - ondemand_active_tests (Exit code 0xc0000135
4428)
4429	 28 - ondemand_dom_api_tests (Exit code 0xc0000135
4430)
4431	 29 - ondemand_error_tests (Exit code 0xc0000135
4432)
4433	 30 - ondemand_key_string_tests (Exit code 0xc0000135
4434)
4435	 31 - ondemand_number_tests (Exit code 0xc0000135
4436)
4437	 32 - ondemand_ordering_tests (Exit code 0xc0000135
4438)
4439	 33 - ondemand_parse_api_tests (Exit code 0xc0000135
4440)
4441	 34 - ondemand_twitter_tests (Exit code 0xc0000135
4442)
4443Errors while running CTest
4444
```

The Exit code seems to indicate that the binary is missing. I can investigate some more if you'd like.

lemire:
This will require an update to our OnDemand documentation, right? If so, I think we should at least create an issue (with a relevant milestone) before merging this PR.

jkeiser:
Oh, I'll fix the docs, thanks.

The error there is new with my last commit, I think--but I'm not sure yet. Debugging.

jkeiser:
Super weird ... Win32 doesn't look any different from the Win64 builds, command line wise. Those executables seemed to get built, too ...

benchmark/kostya/ondemand.h
```diff
@@ -56,10 +56,10 @@ simdjson_really_inline bool OnDemand::Run(const padded_string &json) {
   count = 0;
 
   auto doc = parser.iterate(json);
-  for (ondemand::object coord : doc["coordinates"]) {
-    sum.x += double(coord["x"]);
-    sum.y += double(coord["y"]);
-    sum.z += double(coord["z"]);
+  for (ondemand::object coord : doc.find_field("coordinates")) {
```
lemire, line 59: @jkeiser These are just direct semantic transposition of the previous code, but should we not also include a version with the order-insensitive code... because that's where the performance differential should appear... No? (I think that the numbers would be great to have.)

jkeiser: @lemire I agree, but I'd like to make a benchmark that includes *input* with different order as well, to match. Having more benchmarks is going to be an important part of a release.

jkeiser:
Documentation update added.

benchmark/kostya/ondemand.h
jkeiser: While I *do* want to do that, I went ahead and added the unordered version for comparison purposes. Unordered still seems to be slightly worse than ordered, by roughly 2%, and it is a real difference caused almost entirely by increased instruction count:

| Benchmark | Throughput | Instructions | Branch Misses |
|---|---|---|---|
| LargeRandom<OnDemand> | 634.066 MB/s | 614,119,494 | 972,480 |
| LargeRandom<OnDemandUnordered> | 650.795 MB/s | 636,119,541 | 956,152 |

jkeiser:
Win32 is still failing, and I don't understand it yet.

doc/ondemand.md
```diff
@@ -714,26 +714,16 @@ We expect that the On Demand approach has many of the performance benefits of th
 
 ### Limitations of the On Demand Approach
 
-The On Demand approach has  some limitations:
+The On Demand approach has some limitations:
 
-* Because it operates in streaming mode, you only have access to the current element in the JSON document. Furthermore, the document is traversed in order so the code is sensitive to the order of the JSON nodes in the same manner as an event-based approach (e.g., SAX).
+* Because it operates in streaming mode, you only have access to the current element in the JSON document. Furthermore, the document is traversed in order so the code is sensitive to the order of the JSON nodes in the same manner as an event-based approach (e.g., SAX). (The one exception to this is field lookup, which is more *performant* when the order of lookups matches the order of fields in the document, but which will still work with out-of-order fields, with a performance hit.)
```
lemire, line 719: Ok, so find_field is not going to be in our documentation for the time being, correct? (Just checking.)

jkeiser:
Yeah, the document at present is more of a design doc than a how to.

lemire:
I am building for Win32 and I am getting errors like this...

Screen Shot 2020-12-23 at 11 04 04 AM

lemire:
I must say that in days like today, I am really missing old school makefiles. It is really hard to understand what is going on with CMake.

jkeiser:
Yeah. I can see it *building* some of those--I think--but maybe the fact that it's down in a sub-sub-directory is making it angry?

jkeiser:
(But then I don't understand why it's only Win32 and not all Windows.)

jkeiser:
Based on appveyor history, I think [this](https://github.com/simdjson/simdjson/pull/1351/commits/d0bccf214d56bbe3b3396d7b12c1f6c4d6ecc3de) is the checkin that introduced the issue. The prior commit passed the Win32 test.

lemire:
> (But then I don't understand why it's only Win32 and not all Windows.)

There are more differences than just Win32. It could be that some other flag is the issue.

lemire:
I am getting that the DLL is not found, but only for these ondemand tests...

jkeiser:
Testing the sub-sub-directory theory by having everything build from the parent directory ...

lemire:
I think I see a difference. The normal test directories contain a copy of the dll, the ondemand directories do not.

Screen Shot 2020-12-23 at 11 31 49 AM
Screen Shot 2020-12-23 at 11 31 37 AM

lemire:
That's it. If I manually copy the DLL in the ondemand directory, everything works.

lemire:
I don't think it is about Win32 at all, I think it is about static vs dynamic.

lemire:
See this part...
```
# Copy the simdjson dll into the tests directory
if(MSVC)
  add_custom_command(TARGET basictests POST_BUILD        # Adds a post-build event
    COMMAND ${CMAKE_COMMAND} -E copy_if_different  # which executes "cmake -E copy_if_different..."
        "$<TARGET_FILE:simdjson>"      # <--this is in-file
        "$<TARGET_FILE_DIR:basictests>")                 # <--this is out-file path
endif()
```

You are missing it in the ondemand test directory.

I think that's all there is to it.

jkeiser:
Eww! Also, thank you!

lemire:
@jkeiser In theory, there might be a nicer way than to manually copy the DLL but it works and it is easy to debug and understand.

jkeiser:
@lemire I'm hoping there is a way to attach it to the link_libraries(simdjson) call so that no one ever has to do it.

jkeiser:
But I'm not gonna solve that one right now :)

jkeiser:
OK, let's see if that makes the problem go away!

lemire:
So we really should rename "Win32" by "Win32 dynamic" because it tests both things at once.

We don't test the full matrix, since it is too expensive.

lemire:
@jkeiser We now have something else...

```
61
  CMake Error at tests/ondemand/CMakeLists.txt:21 (add_custom_command):
62
  Error:   No TARGET 'basictests' has been created in this directory.
```

jkeiser:
@lemire tested locally, and should be fixed. Thank you so much for taking the time to investigate!

jkeiser:
Note: I continue to have *some* trepidation about making this change; in particular, I think it will not mesh well with streamed parsing support, since you have to be able to go backwards. Either we'll want to have separate, special streamed parsing ondemand (very very possible we'll want this); or we'll want to introduce something like `object.get_unordered_object()` which will "pin" the buffer instead of unloading it.

Thinking about it, having separate non-streamed parsing actually might make that API even easier to use and safer--we could make a lot of things work that didn't work before ...

jkeiser:
I'm OK with this patch, however, because I am quite certain the ordered behavior is going to confuse users and cause nearly invisible bugs (the kind that happen in production but never in test).

lemire:
@jkeiser I have created a related issue...

https://github.com/simdjson/simdjson/issues/1356

lemire:
I do think that this PR is a great move. It removes a sharp edge to the API.

It creates a puzzle for chunked indexing but not one that is insane.

jkeiser:
Ok, all platforms are green!

lemire:
@jkeiser After we merge, we should send a PR or a warning to people who benchmark our code using the tip of our main branch since we are changing the semantics.

The release notes will also need to be quite explicit about this change... Anyone who is using OnDemand will need to be made aware of the change. (It is fine, but we need to be mindful fo the need for people to update their code.)

jkeiser:
@lemire yeah. I get the impression it hasn't really been used (or used much), but it's worth doing that.

I still don't want to state that the API is finished until we have a streaming version, because (as noted in the issue you filed) I think it may well affect the API. I'd like to try and avoid having a split API (forward and non-forward versions), and there may well be ways to do that, but it will likely affect the API a little bit.

In particular, if we can solve this streaming + unordered field problem in a reasonable way, I'd love to apply that to other values as well, so that you *can* make things like `std::pair<double, double> point = make_pair(object["x"], object["y"])` work. I ran into that problem quite naturally while simplifying examples. (It doesn't now because that creates a `std::pair<value, value>` which then converts both values to double afterwards.) I can imagine ways to do this, but I'm unsure how it'll work out in the end.

So next is either streaming supportability, or real-use-case benchmarks and usage docs. I'm inclined towards the latter, because it makes what we have truly releasable (though still in alpha, as I don't want to imply the API is final until we work out the streaming problem). document_stream and element printing, though easy, just don't seem as important. (Also, element printing is pretty likely to need some of the "go backwards" functionality, since people will often want to first print a value for logging / debugging and then use it.)

I'm going to merge shortly, unless there are objections--I don't see an Approved, but I think you basically have at this point :)

lemire:
(I think I did approve though maybe it did not register.)

PR #1364, opened 2021-01-03
Debug only test option

friendlyanon:
This is in reference to https://github.com/simdjson/simdjson/pull/1352#issuecomment-750321197

Adds a `DEBUG_ONLY` option argument to `add_cpp_test`, which will just turn the function into a noop if the build system generator is a single configuration generator and `CMAKE_BUILD_TYPE` is not set to `Debug` and when the build system generator is multi-config, then will constrain the source set using a `$<CONFIG:Debug>` generator expression, but also add an empty cpp source for every other configuration, so the compiled binary is minimal, and also sets the `add_test(... CONFIGURATIONS Debug)` argument, so the test only runs in debug configuration.

@jkeiser Since you added the test that got fixed with #1352 I would like to ask you on what to do with CI and labels. CI would need some changes how CMake is invoked to properly cover all cases:
```bash
cmake -S <source> -B <build> -D CMAKE_BUILD_TYPE=$BUILD_TYPE ...
cmake --build <build> --config $BUILD_TYPE ...
ctest -C $BUILD_TYPE ...
```
In the above snippet a build type argument is present at every step, so it doesn't matter what build system is actually in use. Consistent invocation would mean that the necessary build type information is always present instead of relying on some potentially wrong default.

### DIFF
--- a/cmake/add_cpp_test.cmake
+++ b/cmake/add_cpp_test.cmake
@@ -1,38 +1,69 @@
 # Helper so we don't have to repeat ourselves so much
-# Usage: add_cpp_test(testname [COMPILE_ONLY] [SOURCES a.cpp b.cpp ...] [LABELS acceptance per_implementation ...])
-# SOURCES defaults to testname.cpp if not specified.
+# Usage: add_cpp_test(<testname>
+#                     [COMPILE_ONLY]
+#                     [DEBUG_ONLY]
+#                     [WILL_FAIL]
+#                     [LIBRARY]
+#                     [SOURCES a.cpp b.cpp ...]
+#                     [LABELS acceptance per_implementation ...]
+#                     [DEPENDENCY_OF ...])
+# SOURCES defaults to <testname>.cpp if not specified.
 function(add_cpp_test TEST_NAME)
   # Parse arguments
-  cmake_parse_arguments(PARSE_ARGV 1 ARGS "COMPILE_ONLY;LIBRARY;WILL_FAIL" "" "SOURCES;LABELS;DEPENDENCY_OF")
-  if (NOT ARGS_SOURCES)
+  cmake_parse_arguments(PARSE_ARGV 1 ARGS "COMPILE_ONLY;LIBRARY;WILL_FAIL;DEBUG_ONLY" "" "SOURCES;LABELS;DEPENDENCY_OF")
+
+  # Is generator multi-config?
+  get_cmake_property(is_multi_config GENERATOR_IS_MULTI_CONFIG)
+
+  # Return if debug only test
+  if(NOT is_multi_config AND ARGS_DEBUG_ONLY AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+    return()
+  endif()
+
+  # Set configurations
+  set(config_args)
+  if(is_multi_config AND ARGS_DEBUG_ONLY)
+    set(config_args CONFIGURATIONS Debug)
+  endif()
+
+  if(NOT ARGS_SOURCES)
     list(APPEND ARGS_SOURCES ${TEST_NAME}.cpp)
   endif()
-  if (ARGS_COMPILE_ONLY)
-    list(APPEND ${ARGS_LABELS} compile_only)
+
+  # Wrap sources in a generator expression and provide a dummy source
+  if(is_multi_config AND ARGS_DEBUG_ONLY)
+    list(TRANSFORM ARGS_SOURCES PREPEND "$<$<CONFIG:Debug>:")
+    list(TRANSFORM ARGS_SOURCES APPEND ">")
+    list(PREPEND ARGS_SOURCES "$<$<NOT:$<CONFIG:Debug>>:${simdjson_SOURCE_DIR}/cmake/empty.cpp>")
+  endif()
+
+  if(ARGS_COMPILE_ONLY)
+    list(APPEND ARGS_LABELS compile)
   endif()
 
   # Add the compile target
-  if (ARGS_LIBRARY)
+  if(ARGS_LIBRARY)
     add_library(${TEST_NAME} STATIC ${ARGS_SOURCES})
-  else(ARGS_LIBRARY)
+  else()
     add_executable(${TEST_NAME} ${ARGS_SOURCES})
-  endif(ARGS_LIBRARY)
+  endif()
 
   # Add test
-  if (ARGS_COMPILE_ONLY OR ARGS_LIBRARY)
+  if(ARGS_COMPILE_ONLY OR ARGS_LIBRARY)
     add_test(
       NAME ${TEST_NAME}
       COMMAND ${CMAKE_COMMAND} --build . --target ${TEST_NAME} --config $<CONFIGURATION>
       WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+      ${config_args}
     )
     set_target_properties(${TEST_NAME} PROPERTIES EXCLUDE_FROM_ALL TRUE EXCLUDE_FROM_DEFAULT_BUILD TRUE)
   else()
-    add_test(${TEST_NAME} ${TEST_NAME})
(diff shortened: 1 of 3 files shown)


cmake/add_cpp_test.cmake
```diff
@@ -41,13 +72,13 @@ function(add_cpp_test TEST_NAME)
   endif()
 
   # Add as a dependency of given targets
-  foreach(dependency_of ${ARGS_DEPENDENCY_OF})
+  foreach(dependency_of IN LISTS ARGS_DEPENDENCY_OF)
     if (NOT TARGET ${dependency_of}_tests)
       add_custom_target(${dependency_of}_tests)
       add_dependencies(all_tests ${dependency_of}_tests)
-    endif(NOT TARGET ${dependency_of}_tests)
+    endif()
```
jkeiser, line 79: Is it recommended that we not put stuff in the endif()?

cmake/add_cpp_test.cmake
```diff
@@ -1,38 +1,69 @@
...
     add_library(${TEST_NAME} STATIC ${ARGS_SOURCES})
-  else(ARGS_LIBRARY)
+  else()
     add_executable(${TEST_NAME} ${ARGS_SOURCES})
-  endif(ARGS_LIBRARY)
+  endif()
 
   # Add test
-  if (ARGS_COMPILE_ONLY OR ARGS_LIBRARY)
+  if(ARGS_COMPILE_ONLY OR ARGS_LIBRARY)
```
jkeiser, line 52: I take it no space after the if() is considered idiomatic for cmake?

jkeiser:
> This is in reference to [#1352 (comment)](https://github.com/simdjson/simdjson/pull/1352#issuecomment-750321197)
> 
> Adds a `DEBUG_ONLY` option argument to `add_cpp_test`, which will just turn the function into a noop if the build system generator is a single configuration generator and `CMAKE_BUILD_TYPE` is not set to `Debug` and when the build system generator is multi-config, then will constrain the source set using a `$<CONFIG:Debug>` generator expression, but also add an empty cpp source for every other configuration, so the compiled binary is minimal, and also sets the `add_test(... CONFIGURATIONS Debug)` argument, so the test only runs in debug configuration.

A couple of things:

1. What if the user is using a release configuration with assertions enabled (i.e. without -DNDEBUG)? Is this a thing much? I've done it before by modifying the release flags, but I don't know for sure if that's the right way.
2. Is it possible to put this in an if() statement in the caller rather than as part of add_cpp_test? It feels independent. I'm not unbendingly opposed to putting it in the function, but it seems like this is not used enough (and possibly *won't* be used enough) to justify adding more flags to the already flagged-up interface of add_cpp_test (which is 100% my fault :).

> 
> @jkeiser Since you added the test that got fixed with #1352 I would like to ask you on what to do with CI and labels. CI would need some changes how CMake is invoked to properly cover all cases:
> 
> ```shell
> cmake -S <source> -B <build> -D CMAKE_BUILD_TYPE=$BUILD_TYPE ...
> cmake --build <build> --config $BUILD_TYPE ...
> ctest -C $BUILD_TYPE ...
> ```
> 
> In the above snippet a build type argument is present at every step, so it doesn't matter what build system is actually in use. Consistent invocation would mean that the necessary build type information is always present instead of relying on some potentially wrong default.

Seems reasonable to me.

Is the name "Debug" meaningful? Do people make other kinds of Debug configurations, and would we need to care about that, or do we own the configuration names?

cmake/add_cpp_test.cmake
friendlyanon: It probably was required in some early CMake versions, but it's been optional for the longest time. Proper indentation already makes it easy to tell blocks apart and if the expressions don't match that's an error, it's just better not have anything in the closing command call.

cmake/add_cpp_test.cmake
friendlyanon: `if` is "just" a command like any other and the other commands aren't spaced apart either.

friendlyanon:
* Since this is only for tests, the project has (should have) full control over the flags in use. Is compiling with `NDEBUG` define in `Release` mode something that tests might require?
* Well if this is one-off, then I'm not opposed to just laying out the commands manually for this one case.
* Configuration names come from the build system. See [CMAKE_CONFIGURATION_TYPES](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html).

jkeiser:
Actually,  I bet we could test for the presence of NDEBUG the same way we treat for fork and wait, avoiding all of this ...

cmake/add_cpp_test.cmake
```diff
@@ -1,38 +1,69 @@
...
+    set(config_args CONFIGURATIONS Debug)
+  endif()
+
+  if(NOT ARGS_SOURCES)
     list(APPEND ARGS_SOURCES ${TEST_NAME}.cpp)
   endif()
-  if (ARGS_COMPILE_ONLY)
-    list(APPEND ${ARGS_LABELS} compile_only)
+
+  # Wrap sources in a generator expression and provide a dummy source
```
jkeiser, line 33: Oh, now I understand it. The distance between this statement and the setting of "is_multi_config" meant I didn't realize these were related. Perhaps if this if() statement and the previous one were both in the same place, this would be easier to grok?

A comment explaining the purpose is still probably warranted.

jkeiser:
OK ... I don't want to slow down your life by redesigning the thing you designed, and I can live with the DEBUG_ONLY option. I see now why you can't use an if() statement externally--if it's multi config.

I don't fully understand the need to add --config arguments every time we run cmake / ctest. I do worry a tiny bit about not testing cmake with default arguments, since that's what users will typically do, but if it's something everyone should just do all the time, then we should document it in the README.

I *am* starting to wonder if we should do this at all, though ... my comment in the original issue may have been Just Wrong. It turns out we can disable individual tests based on the presence of NDEBUG, after the refactor I did, so we might want to compile the executable every time and let the tests decide. If that's the case, I'm sorry I led you down the rabbit hole :(

friendlyanon:
Oh no, don't you worry about my time. I rather spend more time on a solution that benefits *you* better, since the point of changes to CMake files would be to make your (the people actually developing the library) life easier. I fully understand that there is a very steep learning curve to CMake and all the multi-faceted problems it has to and does solve cover a huge ground, so I am happy to help fixing related problems. I can even understand the frustration regarding CMake in a recent PR's discussion, it can indeed get messy quick without the proper knowledge.

Passing config with `--config` and `-C` arguments is required for multi-config generators (XCode, VS, Ninja multi-config, etc), because the build type is determined at build time, not configuration time. "Default" build type in all cases is determined by the underlying build system, which is usually the Debug build type. I wouldn't worry about it too much, the need to build in Release mode will always require (build system dependant) extra parameters.

There wasn't really a rabbit hole to go down, unless you consider playing with CMake and learning it one :)  
The solution in this PR is a pretty straightforward change, but obviously requires some domain knowledge. Being a draft, I merely presented something to discuss and think about.

cmake/add_cpp_test.cmake
friendlyanon: I was hoping the condition coupled with the comment was enough. This block basically makes sure that in configurations other than Debug, the only source this target will compile is the empty cpp file. There is no way to disable targets cleanly based on configuration, so just using a dummy source to compile instead of the actual source(s) was the best option.

cmake/add_cpp_test.cmake
jkeiser: I felt like the end bracket stuff made for nice commentary, so you could tell what was being closed when the opening line is far away. Guess I am just wondering if there is a common style (if you don't know, it's cool).

cmake/add_cpp_test.cmake
jkeiser: I understand; just not sure what "most people" do in cmake :)

jkeiser:
I think we've concluded this isn't necessary anymore?

jkeiser:
I'm going to close, but please absolutely reopen or comment if you think it should be in!

Issue #1361, opened 2020-12-30
Faster bulk loading and total streaming for large machine generated JSON


codewitch-honey-crisis:
I'm building a totally demand streaming parser - everything but field names can be streamed, and it can all run comfortably in 4kB of ram if you want it to. It runs on 8-bit arduinos. It screams on PCs.

Article: https://www.codeproject.com/Articles/5290815/JSON-by-Design-A-Walkthrough-of-a-High-Performance
Latest code: https://github.com/codewitch-honey-crisis/JSON-CPP

For a thorough description of what it is and how it works, please see the article.

It has a fast navigation/extraction mechanism rather than an in-memory model, although you can use in-memory trees with it, it's not a primary feature.

I think I can make simdjson a lot faster in certain situations - like where you don't need full validation. If you're bulk scanning machine generated dumps it doesn't make sense to run a validator off of it to extract data and it wastes lots of cycles.

It would be probably a separate API because it's a pull parser and the closest thing you have is your demand api which accomplishes something similar at a high level, but works entirely differently than what i devised?

My code footprint is small, so it might be worthwhile to create a second API with it for bulkloading?

I can provide my code as a proof of concept right now but the extraction portion currently has a significant bug I'm in the middle of running down.

I'd add SIMD but frankly, I just don't have the skills for that. This is (I believe) an algorithmic improvement over simdjson processing specifically for bulk load scenarios, but I can't fairly test simdjsons parsing/scanning features with mine because simdjson's i/o is so optimized below the parsing that it's just not telling me anything useful.

I'm running an old I5 on a standard western digital black HDD under linux.
1MB blocked fread() on this machine, plus a complement operation on each returned byte runs at about 4000MB/s
This also runs on an arduino at an average of 25kbps on an 8-bit arduino

for optimization i am using no specific SIMD, but I am using strpbrk() over a memory mapped file in some cases, which is pretty fast.

What I think is that this is all portable C++ code, so if it was optimized the way simdjson is it would beat simdjson because it doesn't do validation, if that makes sense. Of course, if you need validation, don't do it this way.

I would interested in contributing if this feature is of interest to anyone.

(i added // notes below)
```
// 200kB file - I've gotten around 700+MB/s on a 20MB file of bulk data but i can't use that for debugging
// targeted value extraction 
Extract episodes - JSONPath equivelent: $..episodes[*].season_number,episode_number,name
	Extracted 112 episodes using 23 bytes of the pool
	Memory: 334.723936MB/s
[...]

PR #1367, opened 2021-01-04
added jsoncpp folder for comparison and possible adaptation/integration

codewitch-honey-crisis:
Added JSON(C++) to jsoncpp for comparison, and initial benchmark. could use help adding certain comparisons, like finding all ids in a document with simdjson along a descendants axis.

It seems to be competitive speed wise, so far, and it uses a lot less RAM than simdjson, especially for large documents, since RAM use is not at all linked to document size with JSON(C++).

This pull request is for possible integration, exploring of alternatives for high speed parsing when bulk loading.

It can stream everything but field names, and it preserves numbers of arbitrary precision for roundtripping. 

It can operate on systems with 4kB of RAM, and 8-bit processors, but scales to modern architectures and virtual memory, so far pretty competitive with SIMD json on my very initial benchmarks without using much RAM, and total streaming for bulk loading. I'm working on more benchmarks now, but i need to make them for bulk data, because simdjson probably beats the pants of my parser for small data.
```
simdjson is using: westmere
simdjson took 46.874 ms to find Canceled
JSON(C++) took 30.66 ms to find Canceled
simdjson took 49.563 ms to extract episode data
JSON(C++) took 56.497 ms to extract episode data

```
This is against the ondemand api over a 20MB pretty printed JSON file.

I'll add more comparisons once i figure out how to use ondemand to search across each different axis (forward, descendants, siblings) - right now i can only get it to do siblings, but it's not fair to make it walk the tree in the code if it doesn't have to, when mine doesn't. Nor do i want to write skewed benchmarks just because i am still learning the api.

See the discussion here: https://github.com/simdjson/simdjson/issues/1361
And a high level overview of JSON(C++) here: https://github.com/codewitch-honey-crisis/JSON-CPP

under jsoncpp there is a main.cpp which will compile under linux with the settings in ./run.sh

## some observations
1. It might be better to skip total validation for machine generated bulk json
2. Stage 1 analysis may not be the most appropriate thing to use for bulk loading
3. It could be worth it to have a separate api like like bulk:: rather than try to integrate with ondemand since ondemand still uses stage 1 and still has memory requirements that are linked to document size, AFAIK
4. I think both libraries need JSONPath support, and i can easily add it to simdjson to use ondemand i think, or if nothing else, the DOM. My library needs more work for that to happen, but i'm going to tackle that at some point, hopefully soon. I say this because my library can be difficult to use. Basically I'm still exposing the guts of a query engine I haven't totally written yet. 

As far as the query engine, I intend to support the forward only subset of JSONPath. That will make my library as easy to use as anything out there. I'll probably make it by way of a code generator that writes code to use the reader. That way it will work on arduinos without more overhead of a query parser and evaluator. I may add an optional library that supports it at runtime. It will work SAX like by giving you callbacks with results.

These are initial thoughts on the advantages and barriers of integrating with simdjson.

...
Adding, this is failing: continuous-integration/travis-ci/pr — The Travis CI build could not complete due to an error

Does anyone know why? Maybe it's trying to build my code? I haven't touched it.

ETA: Sorry for the initial pull request snafu. this was my first pull request ever and I'm still figuring it out. reading the docs only got me so far, apparently


lemire:
I changed the base to development/honey. This PR could be merged safely there.

codewitch-honey-crisis:
Requests: 

For anyone kind enough:

Platform support: I'd really love to see how this performs on windows machines but mine is down, and will remain down until i buy my new machine this month. The library (not main.cpp though) should compile on apple but i need some help implementing map_file() and unmap_file() to use Apple's virtual memory to map a file to the process address space. Without it, it just won't have the speed to be competitive. I can probably implement it, but i have no way of testing what i wrote, so it seems reckless for me to do it.

I need to know how to search along an axis (namely descendants, and forward) in simdjson to write competitive benchmarks. Right now I don't know how to find a field with a particular name anywhere in the document with simdjson. This sort of operation is used both for fast skips, and for document indexing (not like simdjson's indexing, but high level indexing like finding out if the json dump has any new id values compared to last week's dump)

I can't run these types of queries fairly against simdjson until I can use the api in simdjson effectively.

jkeiser:
Does the input have arbitrarily nested objects,  or are the ids all at the same level?

If it's nested at unknown levels,  you'll need to recurse. Simdjson doesn't natively have the method you're talking about.

codewitch-honey-crisis:
It's nested at unknown levels. My implementation never recurses except when
it's creating in memory trees.

It keeps an object depth number, but no stack.

Because of this, validation requires additional code, due to an ambiguity
in the JSON grammar

[ "foo", "bar", "fubar":"baz", "foobar" ]

will skip through (but fail if you actually try to parse it). you either
need to use a stack outside of it if you want to check for that or call
parseSubtree() on the entire document to parse it into memory in order
validate it. This was never intended to fully validate.

That's good to know that simdjson doesn't have that method, but it's
unfortunate because a great use of simdjson would be id indexing out of
large json dumps, i think. (useful for things like checking for updates
from mongoDB repos)

codewitch-honey-crisis:
Other than code size, and some of the things we already covered one of the reasons my code performs the way it does is because most of it is driven my skipping, again using strpbrk() (SIMD optimized, AFAIK) over a mem mapped file pointer. 

Even when you extract, basically what an extraction is is a set of precoded "moves" the reader does for you. About the only wizardry it does there is when you search for multiple fields off an object, it does some creative things to compare all the requested field names with the field under the cursor at the same time so it doesn't scan the same string more than once per field. But it's always skipping through. It doesn't care about structural analysis. It looks for markers in the document.

The only thing it's consistently respectful of structure about is tracking object depth (mostly because it has to in order to search descendants effectively) - which means it always stops what it's doing when it finds '{' or '}' (and often '[' or ']' but I'm trying to remove those checks where i can to speed it up)

For example, here we see it using skipToAny() - skipToAny() is easily the fastest way to move through the document. All it does is keep advancing until it hits a particular single-byte character (all json markers are single byte and i exploit that) and then it returns that character. It examines the current character under the cursor, and when complete advances past the matched character it returns. The fastest implementation of skipToAny() is in MemoryMappedLexSource<TCapacity> but it can be implemented on other sources, like for buffered TCP I/O for example
```
bool skipObjectPart(int depth=1)
{
    clearError();
    // TODO: this is so much faster when i don't have to track arrays
    // for some reason though, that breaks extract() and/or parseSubtree()
    char ch = m_lc.skipToAny("\"{}[]");
    while(0!=ch) {
        switch(ch) {
            case '\"':
                if(!skipString())
                    return false;
                break;

            case '{':
                ++m_objectDepth;
                ++depth;
                if (!m_lc.advance()) {
                    if(m_lc.hasError()) {
                        error(m_lc);
                        return false;
                    }
                    JSON_ERROR(UNTERMINATED_OBJECT);
                    m_state=Error; // unrecoverable
                    return false;
                }
            break;

            case ']':
                --depth;
                if (!m_lc.advance()) {
                    if(m_lc.hasError()) {
                        error(m_lc);
                        return false;
                    }
                    JSON_ERROR(UNTERMINATED_ARRAY);
                    m_state=Error; // unrecoverable
                    return false;
                }
                break;
            case '[':
                ++depth;
                if (!m_lc.advance()) {
                    if(m_lc.hasError()) {
                        error(m_lc);
                        return false;
                    }
                    JSON_ERROR(UNTERMINATED_ARRAY);
                    m_state=Error; // unrecoverable
                    return false;
                }
                break;

            case '}':
                --depth;
                --m_objectDepth;
                if (0>=depth)
                {
                    m_state=(m_lc.eof())?EndDocument:EndObject;
                    if(!m_lc.advance()) {
                        if(m_lc.hasError()) {
                            error(m_lc);
                            return false;
                        }
                    }
                    if(!JsonUtility::skipWhiteSpace(m_lc)) {
                        error(m_lc);
                        return false;
                    }
                    return true;
                }
                if(!m_lc.advance()) {
                    if(m_lc.hasError()) {
                        error(m_lc);
                        return false;
                    }
                }

                break;
        }

        ch=m_lc.skipToAny("\"{}[]");
    }

    if(m_lc.hasError()) {
        error(m_lc);
        return false;

    }
    return false;
}
```

jkeiser:
@codewitch-honey-crisis the benchmarks in master have improved significantly. Look at benchmarks/bench_ondemand.cpp to see the variety of benchmarks and implementations, and add jsoncpp implementations! That'll give you a really concrete idea of where things stand.

jkeiser:
I think it's probably best not to include a comparative benchmark until jsoncpp is released somewhere. Regardless, it's totally cool to have a branch or fork where you're working on this and running the benchmarks to measure up.

Note: once it is released, it'll be better to pull this dependency from github (see [dependencies/CMakeLists.txt](https://github.com/simdjson/simdjson/blob/master/dependencies/CMakeLists.txt#L44-L47) for examples) rather than including the jsoncpp source here permanently.

lemire:
Merging into a separate branch as discussed.

PR #1484, opened 2021-03-05
[draft] Move ondemand public API into the simdjson namespace

jkeiser:
This is what it looks like when you implement #1482.


jkeiser:
Closing. We're going to leave things as-is for 1.0, and revisit when we have finished design of some other important simdjson features that might affect it.

PR #2027, opened 2023-07-05
Move numberparsing functions to architecture top level

jkeiser:
Since number parsing is outside of ondemand at generic/numberparsing.h (which is arguably correct), we should keep its stuff out of the ondemand namespace. We do have documentation (and possible code out in the wild) that relies on `number_type` being in ondemand, so we use a `using` statement to alias that type into ondemand.

Noticed this when I tried to make a copy of ondemand for an experiment and couldn't access number functions anymore :)

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,2 @@
+CompileFlags:
+    CompilationDatabase: build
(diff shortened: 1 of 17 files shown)


lemire (approved):
Uncontroversial, merge away!

Issue #1508, opened 2021-03-20
Consider ARMv7 support


jkeiser:
Currently we only support ARMv8. @easyaspi314 submitted a PR for ARMv7, but we need to figure out how to:

* Test it in CI (or come up with a reasonable way to have an "unsupported architecture"--perhaps through #defines)
* Check its performance (though if it's in the unsupported bucket, that would be fine)

Original PR: https://github.com/simdjson/simdjson/pull/330

PR #2040, opened 2023-07-20
Peek(0) is an error on 32-bit systems, where pointers are indistingui…

calumapplepie:
…shable from int32_t

yeah, that.  I know 32 bit systems aren't really supported, especially since #1508 isn't done, but having a working 32-bit version helps in my project, SilliconJson.  Since the message that prints if you do use a 32 bit system doesn't say "this won't work at all", lets keep it somewhat working.

### DIFF
--- a/include/simdjson/generic/ondemand/json_iterator-inl.h
+++ b/include/simdjson/generic/ondemand/json_iterator-inl.h
@@ -254,7 +254,7 @@ simdjson_inline const uint8_t *json_iterator::return_current_and_advance() noexc
 
 simdjson_inline const uint8_t *json_iterator::unsafe_pointer() const noexcept {
   // deliberately done without safety guard:
-  return token.peek(0);
+  return token.peek();
 }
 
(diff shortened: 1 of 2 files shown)


lemire:
@calumapplepie This change is fine and we will merge, but I fail to see how it is related to 32-bit systems.

lemire:
Note that we do build, in CI, under several 32-bit systems (and also big endian IBM systems).

calumapplepie:
All that I know is that when building for a Cortex A9 using Vitis (GCC-12, though I didn’t verify with an upstream build) the build failed with complaints about ambiguity between peek(int) and peek(pointer)Sent from my iPhoneOn Jul 20, 2023, at 8:09 PM, Daniel Lemire ***@***.***> wrote:﻿
Note that we do build, in CI, under several 32-bit systems (and also big endian IBM systems).

lemire:
I'll merge the PR, because it is a code simplification.

PR #2242, opened 2024-08-29
add pdep/pext compress implementation

Validark:
Had this idea, but unfortunately I can't get it to compile so I'm a bit stuck.

I get a couple compile errors that look like this:

```
In file included from /usr/lib/gcc/x86_64-linux-gnu/11/include/x86gprintrin.h:43,
                 from /usr/lib/gcc/x86_64-linux-gnu/11/include/x86intrin.h:27,
                 from /home/niles/Documents/github/simdjson/include/simdjson/haswell/intrinsics.h:12,
                 from /home/niles/Documents/github/simdjson/include/simdjson/haswell/begin.h:4,
                 from /home/niles/Documents/github/simdjson/include/simdjson/haswell.h:4,
                 from /home/niles/Documents/github/simdjson/src/haswell.cpp:8,
                 from /home/niles/Documents/github/simdjson/src/simdjson.cpp:27:
/usr/lib/gcc/x86_64-linux-gnu/11/include/bmi2intrin.h: In member function ‘void simdjson::haswell::{anonymous}::simd::base8_numeric<T>::compress(uint32_t, L*) const [with L = unsigned char; T = unsigned char]’:
/usr/lib/gcc/x86_64-linux-gnu/11/include/bmi2intrin.h:76:1: error: inlining failed in call to ‘always_inline’ ‘long long unsigned int _pext_u64(long long unsigned int, long long unsigned int)’: target specific option mismatch
   76 | _pext_u64 (unsigned long long __X, unsigned long long __Y)
      | ^~~~~~~~~
In file included from /home/niles/Documents/github/simdjson/include/simdjson/haswell/begin.h:13,
                 from /home/niles/Documents/github/simdjson/include/simdjson/haswell.h:4,
                 from /home/niles/Documents/github/simdjson/src/haswell.cpp:8,
                 from /home/niles/Documents/github/simdjson/src/simdjson.cpp:27:
/home/niles/Documents/github/simdjson/include/simdjson/haswell/simd.h:139:39: note: called from here
  139 |   uint64_t wanted_indices2 = _pext_u64(iota, expanded_mask2);
      |                              ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

1. How do I get this to compile?
2. What is the right way to implement this feature? Should `HAS_FAST_PDEP_AND_PEXT` become runtime-dispatched code?

### DIFF
--- a/include/simdjson/haswell/simd.h
+++ b/include/simdjson/haswell/simd.h
@@ -125,6 +125,30 @@ namespace simd {
     // sensible, but the AVX ISA makes this kind of approach difficult.
     template<typename L>
     simdjson_inline void compress(uint32_t mask, L * output) const {
+#define HAS_FAST_PDEP_AND_PEXT 1
+#if HAS_FAST_PDEP_AND_PEXT == 1
+  const uint64_t nibble_msk = 0x1111111111111111;
+  const uint64_t iota = 0xFEDCBA9876543210;
+
+  // duplicate each bit into a nibble of the same value
+  uint64_t expanded_mask1 = _pdep_u64(~mask, nibble_msk) * 0xF;
+  uint64_t expanded_mask2 = _pdep_u64((~mask) >> 16, nibble_msk) * 0xF;
+
+  // Simulate a 16-wide compaction via `pext`, producing up to 16 nibbles
+  uint64_t wanted_indices1 = _pext_u64(iota, expanded_mask1);
+  uint64_t wanted_indices2 = _pext_u64(iota, expanded_mask2);
+
+  // Move to a vector
+  __m128i nibble_vec1 = _mm_cvtsi64_si128(wanted_indices1);
+  __m128i nibble_vec2 = _mm_cvtsi64_si128(wanted_indices2);
+
+  // Spread each nibble to a byte
+  __m128i byte_vec1 = _mm_and_si128(_mm_unpacklo_epi8(nibble_vec1, _mm_srli_epi16(nibble_vec1, 4)), _mm_set1_epi32(0x0F0F0F0F));
+  __m128i byte_vec2 = _mm_and_si128(_mm_unpacklo_epi8(nibble_vec2, _mm_srli_epi16(nibble_vec2, 4)), _mm_set1_epi32(0x0F0F0F0F));
+
+  _mm_storeu_si128( reinterpret_cast<__m128i *>(output), _mm_shuffle_epi8(_mm256_extracti128_si256(*this, 0), byte_vec1));
+  _mm_storeu_si128( reinterpret_cast<__m128i *>(output + count_ones(~mask & 0xFFFF)), _mm_shuffle_epi8(_mm256_extracti128_si256(*this, 1), byte_vec2));
+#else // !HAS_FAST_PDEP_AND_PEXT
       using internal::thintable_epi8;
       using internal::BitsSetTable256mul2;
       using internal::pshufb_combine_table;
@@ -167,6 +191,7 @@ namespace simd {
       _mm_storeu_si128( reinterpret_cast<__m128i *>(output), v128);
       v128 = _mm256_extractf128_si256(almostthere, 1);
       _mm_storeu_si128( reinterpret_cast<__m128i *>(output + 16 - count_ones(mask & 0xFFFF)), v128);
+#endif // !HAS_FAST_PDEP_AND_PEXT
     }
 
     template<typename L>


Validark:
Maybe needed? https://stackoverflow.com/questions/466377/how-to-detect-what-cpu-is-being-used-during-runtime

lemire:
We do not currently target BMI2, see this line...

https://github.com/simdjson/simdjson/blob/e341c8b43861b43de29c48ab65f292d997096953/include/simdjson/haswell/begin.h#L7

We do not do so because we don't use BMI2 in the haswell kernel.

Note that if we are going to use pdep/pext, we are going to have to omit pre-Zen3 AMD processors were these instructions are super slow. It may require an entirely new kernel which is a bit complicated.

It is entirely safe to use pdep/pext in the icelake kernel however.

Validark:
Why not target BMI2 for Haswell? Even if you want to avoid PDEP/PEXT because of AMD, don't you still think the other instructions are nice? Does this mean that the compiler can't insert BMI2 instructions at all or does it just mean that you don't use BMI2 intrinsics?

lemire:
> Why not target BMI2 for Haswell? Even if you want to avoid PDEP/PEXT because of AMD, don't you still think the other instructions are nice? Does this mean that the compiler can't insert BMI2 instructions at all or does it just mean that you don't use BMI2 intrinsics?

Currently, GCC and LLVM won't use BMI2 in the haswell kernel. The reason we do not do it is because we do not expect benefits. Of  course, that's an empirical issue. I will investigate later. :-)

lemire:
Note that the initial versions of simdjson used BMI2 (pdep/pext). At least the way we were using it was not clearly beneficial on Intel processors, and, at the time, quite bad for AMD processors. So we progressively removed it.

lemire:
(We do use BMI2 in the icelake kernel.)

Validark:
Well, according to [Agner Fog's instruction tables](https://www.agner.org/optimize/instruction_tables.pdf), `BZHI` has 1 cycle latency and a reciprocal throughput of 0.25 (i.e. you can start up to 4 per cycle if the ports are available) for Zen 1 and Zen 2. For every other machine, it's 1 cycle latency with 0.5 reciprocal throughput. [uops.info](https://uops.info/table.html?search=bzhi&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_ADLP=on&cb_ADLE=on&cb_measurements=on&cb_doc=on&cb_base=on&cb_bmi=on) confirms this, although it includes the relatively recent [Alder Lake, which appears to have regressed its performance to 3 cycles of latency and a throughput of 1](https://uops.info/table.html?search=bzhi&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_ADLP=on&cb_ADLE=on&cb_measurements=on&cb_doc=on&cb_base=on&cb_bmi=on).

###### (I assume you know what reciprocal throughput is, but honestly I forgot which is which sometimes)

`BZHI` can be called directly through an intrinsic [due to its better handling of 64 as a second input](https://github.com/zwegner/zp7/blob/master/zp7.c#L202-L219) (if the 2nd input has a bottom byte of 64-255, i.e. 0x40-0xff, it will return the first input unchanged), or it can be [generated by the compiler](https://c.godbolt.org/z/7z4v73oc7) when the idiom is used, although obviously that code has undefined behavior when `index` is 64 or greater, so for many non-BMI2-optimized builds it would likely truncate the second input to just the lower 6 bits ([not that Clang would warn you about that by default, but don't worry, `zig cc` has better defaults](https://c.godbolt.org/z/sdEnaYT8x)).

The other instructions `MULX`, `RORX`, `SARX`, `SHRX`, `SHLX` are just special versions of normal operations we already had support for, but these don't affect flags and have a separate destination register from the two input operands, so these are potential wins that the compiler can give you for free. (although they can't use immediates)

So it seems to me the only drawback of BMI2 would be pre-Zen3 AMD processors being really slow at `PDEP`/`PEXT`, but I don't think compilers know how to generate those automatically [[1](https://github.com/llvm/llvm-project/issues/72088), [2](https://github.com/llvm/llvm-project/issues/101915), [3](https://zig.godbolt.org/z/fbKajMM5P)], nor would they for pre-Zen3 AMD, if their cost model was correct.

So I don't really see the point of disabling BMI2. Just because the custom `PDEP`/`PEXT` routines you had before didn't work out doesn't mean it's bad to enable BMI2 as an option for compilers.

Validark:
Update: I added `bmi2` to that haswell target like you said. This now compiles. Is it faster? Well, maybe. I can't really tell by just running it on the twitter.json benchmark and doing eyeball statistics ([definition](https://www.youtube.com/watch?v=r-TLSBdHe1A&t=1076s)). However, according to LLVM's cost model, [my version is faster](https://c.godbolt.org/z/q7oMqzT6P) (I am running a Zen 3 machine). It also means we don't have to include 2576 bytes for the 3 tables, so that seems like a win. However, I am not sure how you would prefer to account for the AMD problem. Let me know what you think.

lemire:
@Validark 

> This now compiles. Is it faster? 

Thanks. I will review and assess.


We might introduce a kernel between icelake and haswell if the performance gains are sufficiently clear.

Of course, a Zen 4 or Zen 5 will use the icelake kernel (where we do use pdep and pext).

Validark:
> > So I don't really see the point of disabling BMI2
> 
> It is not enabled by default. So we are not disabling it. However, it is interesting to check whether enabling it would be beneficial, and I plan to investigate. As your analysis suggests, the important instruction is `BZHI`.

My message was an attempt to explain everything that is included in BMI2, not say which instructions are most important to simdjson.

| instruction | occurrences |
|:-:|:-:|
|`BZHI` ||
|||
|||

> 
> > It also means we don't have to include 2576 bytes for the 3 tables,
> 
> We need the tables because we can't assume that BMI2 is present.

Validark:
> > So I don't really see the point of disabling BMI2
> 
> It is not enabled by default. So we are not disabling it. However, it is interesting to check whether enabling it would be beneficial, and I plan to investigate. As your analysis suggests, the important instruction is `BZHI`.

My message was an attempt to explain everything that is included in BMI2, not say which instructions are most important to simdjson. Based on a objdump, here are the number of occurrences for each instruction:

| instruction | occurrences |
|:-:|:-:|
|`PDEP` |0|
|`PEXT` |0|
|`BZHI` |0|
|`MULX`|10|
|`RORX`|0|
|`SARX`|0|
|`SHRX`|12|
|`SHLX`|12|

They aren't used much, but they do appear in simdjson. And they can eliminate unnecessary internal flag register allocation and possibly eliminate an unnecessary `mov` instruction as well. 

> 
> > It also means we don't have to include 2576 bytes for the 3 tables,
> 
> We need the tables because we can't assume that BMI2 is present.

Haswell has BMI2 though, right? So you can assume that BMI2 is present. LLVM gives you BMI2 for the Haswell cpu target.

lemire:
> They aren't used much, but they do appear in simdjson. 

It is an empirical question. By how much does enabling BMI2 in LLVM or GCC improves performance. As I wrote, I plan to review this issue experimentally.

> Haswell has BMI2 though, right? So you can assume that BMI2 is present. 

We do not assume that the processor supports the equivalent of Haswell. This is determined at runtime by CPU detection.

Validark:
> > They aren't used much, but they do appear in simdjson.
> 
> It is an empirical question. By how much does enabling BMI2 in LLVM or GCC improves performance. As I wrote, I plan to review this issue experimentally.

Feel free, but I think you're unlikely to notice a big difference in practice, these are sort of micro-optimizations. Yes, they save the CPU a little work, but it might be able to do that extra work without adding to the runtime.

> 
> > Haswell has BMI2 though, right? So you can assume that BMI2 is present.
> 
> We do not assume that the processor supports the equivalent of Haswell. This is determined at runtime by CPU detection.

Right, but I'm only talking about the Haswell kernel? I added BMI2 to the list of features that Haswell supports, so if we dispatch to the Haswell kernel, that code can use BMI2 instructions.

I am trying to get you to answer the question of why BMI2 is not enabled for the Haswell kernel. You said you would do it if it's beneficial, which you would test empirically. But my question isn't "Should it be enabled? ", my question is, "Why not?" I'm asking what's the harm in saving the CPU a small amount of work or power consumption? I'm basically implying that I think the default position is to have BMI2 enabled for the Haswell kernel, whereas you're saying that the default position is to have BMI2 disabled for the Haswell kernel. I'm questioning why you think that way.

I think that these instructions were added to the CPU as an optimization, so if the compiler agrees, i.e., wants to insert it in your Haswell specific code, why not just let it? Again, I am not sure that the specific circumstances in which those instructions are inserted are sufficient to see a performance improvement, but the potential is there, so why not? I definitely don't think you'll see performance degradation. **Why not** just enable it for the Haswell kernel?

lemire:
@Validark At least under some versions of GCC, and some hardware, enabling BMI2 appears to lower the performance. It does increase slightly the instruction count.

See https://github.com/simdjson/simdjson/pull/2243

For LLVM, it is beneficial so I am currently considering enabling it only under LLVM.

Validark:
> @Validark At least under some versions of GCC, and some hardware, enabling BMI2 appears to lower the performance. It does increase slightly the instruction count.
> 
> See https://github.com/simdjson/simdjson/pull/2243
> 
> For LLVM, it is beneficial so I am currently considering enabling it only under LLVM.

That's sad that GCC does worse with it on. Well, c'est la vie.

lemire:
@Validark 

Even here, in this PR, you can see that our CI reports a performance degradation:

Screenshot 2024-09-03 at 11 23 21 AM

However, I am not giving up. I think that if we are sufficiently careful, we could get more performance in some cases with BMI2.

PR #2327, opened 2025-01-25
fix fuzzing when running on icelake

pauldreik:
fuzzing locally on an icelake system does not work, seems like fallback+haswell+icelake+westmere is one too much to fit in the max capacity of 3.


lemire (approved):

pauldreik:
merged, the test failures are unrelated.

PR #2398, opened 2025-07-19
Auto Convert  (work in progress)

the-moisrex:
Here's an attempt at making a better API for converting stuff.

Started from [here](https://x.com/lemire/status/1946517171838103744).

## Goals

- [ ] Easy deserializing: `Car car = to("json string")`
- [ ] Easy serializing:   `std::string str = to(Car{})`
- [ ] Work with ranges:   `for (Car car : to("json string") | to<Car>()) { ... }`
  - [ ] Or maybe: `for (Car car : to<Car>("...")) { ... }`
  - [ ] Or maybe if we can pull it off: `for (Car car : to("...")) { ... }`

## Challenges

- [ ] `simdjson_result<T>` needs to know what `T` is; thus the need for `auto_parser`.
  - [ ] Maybe something like `simdjson_result<void>` would be a better place for it than `auto_parser`?!
- [ ] `simdjson_result<ondemand::array_iterator>` is not an actual iterator (`std::input_or_output_iterator`); so thus the need for `auto_iterator`.
- [ ] `std::optional<Car> car = to("...")` is not gonna work because the compiler doesn't know if to call the `operator Car()` and put it into `optional` or to call `operator std::optional<Car>` directly.
- [ ] `Car* car = to("...")` won't work because I don't want to malloc anything.

I had more ideas, but since nothing seems to be what they say they are (iterators are not iterators, ...), I have to think more about them.

**This is not complete yet**; I'm putting it here so the authors can see what's going on so far, so I wouldn't have to rollback too much if that's not the way they want things to go.

### DIFF
--- a/tests/ondemand/ondemand_convert_tests.cpp
+++ b/tests/ondemand/ondemand_convert_tests.cpp
@@ -171,6 +171,19 @@ bool to_bad_array() {
   TEST_SUCCEED();
 }
 
+bool test_no_errors() {
+  TEST_START();
+  for (auto val : simdjson::from(json_cars) | simdjson::no_errors) {
+    Car car{};
+    val.get(car);
+    if (car.year < 1998) {
+      std::cerr << car.make << " " << car.model << " " << car.year << std::endl;
+      return false;
+    }
+  }
+  TEST_SUCCEED();
+}
+
 bool to_clean_array() {
   TEST_START();
   for (Car const car : simdjson::from(json_cars) | simdjson::to<Car>) {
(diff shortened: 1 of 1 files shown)


lemire:
@the-moisrex 


Please see

https://github.com/simdjson/simdjson/issues/2382

I think it relates to some of your concerns, maybe.

the-moisrex:
@lemire Any idea when this becomes `NULL`? Does calling `.begin()` multiple times cause this?

image

the-moisrex:
Okay, now it looks like this for arrays:

```cpp
  for (Car const car : from(json_cars) | to<Car>) {
      // ...
  }

  for (auto val : from(json_cars) | no_errors) {
      Car car;
      val.get(car); // no need to check for errors anymore
      // ...
  }
```

Now let's move on to Objects; ideas? Maybe like this:

```cpp
Car car = from("...")["car1"];
```

lemire:
I am going to close this PR to avoid confusion. See https://github.com/simdjson/simdjson/pull/2406

PR #2578, opened 2026-01-04
moving the builder files in their own directory

lemire:
We had the rather unpleasant design where the builder API was mixed up with the ondemand files. This PR splits them up, for more clarity.

Once this is merged, we'll be able to move the SIMD accelerated part (the string escaping) in the various kernel directories which will also be significantly cleaner and scalable.


georgthegreat:
I think this PR broke capability to directtly `#include <simdjson/ondemand.h>`

Here is the compiler error I receive:

```
/home/thegeorg/arcadia/contrib/libs/simdjson/include/simdjson/generic/ondemand/amalgamated.h:2:2: error: simdjson/generic/ondemand/dependencies.h must be included before simdjson/generic/ondemand/amalgamated.h!
    2 | #error simdjson/generic/ondemand/dependencies.h must be included before simdjson/generic/ondemand/amalgamated.h!
```

@lemire is this error related? I am updating from 4.2.4 to 4.3.0

lemire:
@georgthegreat Our public header is `simdjson.h`. You can just use this header...

```
#include <iostream>
#include "simdjson.h"
using namespace simdjson;
int main(void) {
    ondemand::parser parser;
    padded_string json = padded_string::load("twitter.json");
    ondemand::document tweets = parser.iterate(json);
    std::cout << uint64_t(tweets["search_metadata"]["count"]) << " results." << std::endl;
}
```

I am surprised that `#include <simdjson/ondemand.h>` ever worked. This was not documented or tested.

georgthegreat:
Ok, I will fix our code.
We used these includes to speed up compilation.

lemire:
@georgthegreat We could add such a feature (selective compilation) but I would prefer to do it with feature macros (i.e., we only use X and Y and not Z).

georgthegreat:
It would be nice to have an issue flagged for tracking purposes.
Thanks in advance!

lemire:
@georgthegreat See https://github.com/simdjson/simdjson/issues/2615

PR #2587, opened 2026-01-14
feat: support stripping outer array in iterate_many

mariam851:
Hi Daniel (@lemire),

I hope you're doing well.

This PR addresses issue #2252 by enabling `iterate_many` to handle JSON documents wrapped in a top-level array. 

As discussed in the issue, many streaming sources wrap their output in a large array. This change allows `simdjson` to process these documents efficiently without requiring the user to manually strip the brackets or wait for the entire array to be loaded.

### Key Changes:
- **Automatic Array Stripping:** Updated `iterate_many` to skip the leading whitespace and the initial `[` character if present.
- **Improved Flexibility:** The parser now detects the start of the array and automatically sets `allow_comma_separated` to `true` to handle the subsequent objects within the array.
- **Backward Compatibility:** Existing `iterate_many` functionality for NDJSON/multiple concatenated objects remains unaffected.

## Motivation
The goal is to provide a seamless "Out-of-the-box" experience for users dealing with array-wrapped streaming JSON, which is a common pattern in REST APIs.

I would appreciate your feedback on this implementation. I am a big admirer of your work on high-performance parsers and would be honored to contribute more to `simdjson`.

Best regards,
Mariam

### DIFF
--- a/include/simdjson/generic/ondemand/parser-inl.h
+++ b/include/simdjson/generic/ondemand/parser-inl.h
@@ -134,12 +134,25 @@ simdjson_warn_unused simdjson_inline simdjson_result<json_iterator> parser::iter
 }
 
 inline simdjson_result<document_stream> parser::iterate_many(const uint8_t *buf, size_t len, size_t batch_size, bool allow_comma_separated) noexcept {
-  // Warning: no check is done on the buffer padding. We trust the user.
+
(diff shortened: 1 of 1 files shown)


lemire:
I think that this does not work and it might derive from a misunderstanding as to what is sought. A reasonable newline separated JSON input might be the following:

```
[1,2,3]
[1,2,3]
[1,2,3]
[1,2,3]
...
```

PR #2591, opened 2026-01-17
various minor changes to fractured JSON support

lemire:
This is a PR on top of  @FranciscoThiesen's PR regarding  Fractured JSON.

- I am removing two functions that do not appear important.
- I am adding some documentation
- I am tweaking the tests (and other bits) so that they use our API is the 'proper' manner: we avoid `value()` and `error()` as it is error prone.


FranciscoThiesen:
LGTM!

PR #2598, opened 2026-01-29
updating the reflection benchmarks

lemire:
In our main branch, the C++26 reflection benchmarks were outdated (and wrong). This PR merges the changes @FranciscoThiesen's separate branch (francisco/ablation_study) with a few additional fixes:

- The file `benchmark/static_reflect/citm_catalog_benchmark/rapidjson_citm_catalog_data.h` did not build and appeared to be out of sync. Basically, some values have been optionals, and the code did not reflect that. 
- Though it is not part of our public API, one overload to the `simdjson::builder::to_json` function returned a simdjson_error (which is an exception) instead of a `simdjson::error_code` (which is just a number, essentially). This was a bug in the main library (although unlikely to affect anyone at this time since it is not a documented function). Of course, I caught this because because I added checks for the return values (the error).
- I have updated our `p2996/README.md` so that it is clearer.
- I have also switched some simdjson functions from exception-based to exception-less. (This is not very important).

I note that Nlohmann parsing is disabled for the twitter dataset.

Note that @FranciscoThiesen's branch is not quite in sync as far as the main library is concerned, so there are some differences, possibly.

### DIFF
--- a/p2996/README.md
+++ b/p2996/README.md
@@ -54,7 +54,7 @@ Importantly, we build the experimental LLVM compiler based on the current state 
 ```bash
 CXX=clang++ cmake -B buildreflect -D SIMDJSON_STATIC_REFLECTION=ON -DSIMDJSON_DEVELOPER_MODE=ON
 ```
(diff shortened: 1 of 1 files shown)


lemire:
Merging.

PR #2604, opened 2026-02-14
"Add POWER/Altivec support for fast_needs_escaping"

anosvoldi883:
This PR adds POWER processor support for the `fast_needs_escaping` function.

## Implementation Details
- Uses Altivec/VSX intrinsics (POWER's SIMD instruction set)
- Follows the same algorithm as ARM NEON and x64 SSE2 versions
- Checks for: double quotes ("), backslashes (\), and control chars (< 32)
- Processes 16 bytes at a time for efficiency

## Testing
I don't have access to POWER hardware for testing. The implementation follows
the existing ARM/x64 patterns and uses equivalent Altivec intrinsics. Please
let me know if there's CI available for POWER testing or if you can verify
the functionality.

### DIFF
--- a/include/simdjson/generic/builder/json_string_builder-inl.h
+++ b/include/simdjson/generic/builder/json_string_builder-inl.h
@@ -133,6 +133,55 @@ simdjson_inline bool fast_needs_escaping(std::string_view view) {
   }
   return _mm_movemask_epi8(running) != 0;
 }
+#elif defined(__ALTIVEC__) || defined(__VSX__)
+// POWER processor implementation using Altivec/VSX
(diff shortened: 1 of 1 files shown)


lemire:
I am sorry, @FranciscoThiesen optimized the code, showing that instead of fast_needs_escaping, we are better off with  `find_next_json_quotable_character`. See https://github.com/simdjson/simdjson/blob/af2a43610e7c3560a24ab4016b186260dee748b8/include/simdjson/generic/builder/json_string_builder-inl.h#L165

It is a very similar function, but more general.

PR #2605, opened 2026-02-16
perf: SIMD string escaping and batch integer formatting optimizations

FranciscoThiesen:
## Summary

Performance optimizations for the JSON serialization path in the static reflection builder API, achieving significant improvements on string-heavy workloads.

| Benchmark | Baseline | Optimized | Improvement |
|-----------|----------|-----------|-------------|
| Twitter (string-heavy) | 4,330 MB/s | 5,647 MB/s | **+30.4%** |
| CITM Catalog (numeric-heavy) | 2,912 MB/s | 3,086 MB/s | **+6.0%** |

All measurements on ARM64 (Apple Silicon via Docker with p2996 clang).

---

### Tools Used

1. **perf record** with DWARF call graphs for sampling
2. **perf annotate** for instruction-level cycle attribution
3. **perf report** for function-level analysis

### Key Findings

**Twitter benchmark (before optimization):**
```
54.76%  atom<std::string>                      # String serialization dominant
24.85%  find_next_json_quotable_character      # Scalar position finding
10.90%  atom<unsigned long>                    # Integer formatting
```

**CITM benchmark (before optimization):**
```
31.20%  atom<unsigned long>         # Integer formatting dominant  
26.55%  atom<CITMArea>              # Struct with large int arrays
```

---

### Problem

The serialization path performed redundant scanning:
1. `fast_needs_escaping()` - SIMD scan to check IF any quotable character exists
2. `find_next_json_quotable_character()` - Scalar byte-by-byte scan to find WHERE

### Solution

Unified both operations into a single SIMD pass that directly locates the first quotable character:

```cpp
// NEON implementation - finds exact position in one pass
while (remaining >= 16) {
  uint8x16_t word = vld1q_u8(ptr);
  uint8x16_t needs_escape = vceqq_u8(word, v34);  // '"'
  needs_escape = vorrq_u8(needs_escape, vceqq_u8(word, v92));  // '\\'
  needs_escape = vorrq_u8(needs_escape, vcltq_u8(word, v32));  // control chars

  if (vmaxvq_u32(vreinterpretq_u32_u8(needs_escape)) != 0) {
    // Extract position using trailing zero count on 64-bit lanes
    uint64x2_t as64 = vreinterpretq_u64_u8(needs_escape);
    uint64_t lo = vgetq_lane_u64(as64, 0);
    if (lo != 0) {
      return offset + __builtin_ctzll(lo) / 8;
    } else {
      return offset + 8 + __builtin_ctzll(vgetq_lane_u64(as64, 1)) / 8;
    }
  }
  ptr += 16;
  remaining -= 16;
}
```

Modified `write_string_escaped` to use position finding directly:

```cpp
// Before: Two scans
if (!fast_needs_escaping(input)) { ... }
size_t location = find_next_json_quotable_character(input, 0);

// After: Single scan
size_t location = find_next_json_quotable_character(input, 0);
if (location == mysize) {
  memcpy(out, input.data(), input.size());  // Fast path
  return input.size();
}
```

### Architecture Support
- **NEON (ARM64)**: Full SIMD implementation
- **SSE2 (x86-64)**: Full SIMD implementation using `_mm_movemask_epi8` + `__builtin_ctz`
- **Fallback**: Scalar loop for other platforms

---

### Problem

The original integer-to-string conversion processed 2 digits per iteration. Profiling showed **39% of CITM serialization time** spent on the 2-byte store instruction (`sturh` on ARM64). With CITM integers averaging 8.8 digits, this meant 4+ store operations per number.

### Solution

Process 4 digits per iteration, reducing store count and division operations by half:

```cpp
// Before: 2 digits per iteration
while (pv >= 100) {
  memcpy(write_pointer - 1, &decimal_table[(pv % 100) * 2], 2);
  write_pointer -= 2;
  pv /= 100;
}

// After: 4 digits per iteration for large numbers
while (pv >= 10000) {
  unsigned_type q = pv / 10000;
  unsigned_type r = pv % 10000;
  unsigned_type r_hi = r / 100;  // High 2 digits
  unsigned_type r_lo = r % 100;  // Low 2 digits
  memcpy(write_pointer - 1, &decimal_table[r_lo * 2], 2);
  memcpy(write_pointer - 3, &decimal_table[r_hi * 2], 2);
  write_pointer -= 4;
  pv = q;
}
// Falls through to 2-digit loop for remaining 1-4 digits
```

Division by 10000 compiles to efficient multiply-high instruction (`umulh` on ARM64).

---

## Test Plan

- [ ] All existing unit tests pass
- [ ] Benchmark output matches byte-for-byte (JSON correctness)
- [ ] Test on both ARM64 and x86-64 architectures

### DIFF
--- a/include/simdjson/generic/builder/json_string_builder-inl.h
+++ b/include/simdjson/generic/builder/json_string_builder-inl.h
@@ -157,10 +157,9 @@ find_next_json_quotable_character_scalar(const std::string_view view,
   return size_t(view.size());
 }
 
-// SIMD-accelerated position finding eliminates redundant scanning.
-// Previously, write_string_escaped called fast_needs_escaping (SIMD scan to
-// check IF escape needed) then find_next_json_quotable_character (scalar scan
-// to find WHERE). This unified approach does both in one pass.
+// SIMD-accelerated position finding that directly locates the first quotable
+// character, combining detection and position extraction in a single pass to
+// minimize redundant work.
 #if SIMDJSON_EXPERIMENTAL_HAS_NEON
 simdjson_inline size_t
 find_next_json_quotable_character(const std::string_view view,
@@ -292,9 +291,8 @@ SIMDJSON_CONSTEXPR_LAMBDA void escape_json_char(char c, char *&out) {
   }
 }
 
-// Uses SIMD position finding directly instead of separate fast_needs_escaping
-// check followed by scalar position search. This eliminates redundant scanning
-// that was visible in profiling (24.85% of Twitter serialization time).
+// Writes the escaped version of input to out, returning the number of bytes
+// written. Uses SIMD position finding to locate quotable characters efficiently.
 inline size_t write_string_escaped(const std::string_view input, char *out) {
   size_t mysize = input.size();
 
@@ -507,9 +505,6 @@ simdjson_inline void string_builder::append(number_type v) noexcept {
   else SIMDJSON_IF_CONSTEXPR(std::is_unsigned<number_type>::value) {
     // Process 4 digits at a time instead of 2, reducing store operations
     // and divisions by approximately half for large numbers.
-    // Profiling showed 39% of CITM serialization time in the 2-byte store
-    // instruction. With CITM integers averaging 8.8 digits, this meant 4+
-    // stores per number. 4-digit batching cuts this roughly in half.
     constexpr size_t max_number_size = 20;
     if (capacity_check(max_number_size)) {
       using unsigned_type = typename std::make_unsigned<number_type>::type;


lemire:
Verified. Merging.

samyron:
I realize this is closed but I have a question related to `Optimization 1`

```
<snip>
  if (vmaxvq_u32(vreinterpretq_u32_u8(needs_escape)) != 0) {
    // Extract position using trailing zero count on 64-bit lanes
    uint64x2_t as64 = vreinterpretq_u64_u8(needs_escape);
    uint64_t lo = vgetq_lane_u64(as64, 0);
    if (lo != 0) {
      return offset + __builtin_ctzll(lo) / 8;
    } else {
      return offset + 8 + __builtin_ctzll(vgetq_lane_u64(as64, 1)) / 8;
    }
  }
<snip>
```

There is very similar code in [ruby/json](https://github.com/ruby/json/blob/master/ext/json/ext/simd/simd.h#L126-L127). However, we use something similar to (minus where we apply the mask):

```
<snip>
  const uint8x8_t res = vshrn_n_u16(vreinterpretq_u16_u8(needs_escape), 4);
  const uint64_t mask = vget_lane_u64(vreinterpret_u64_u8(res), 0);
  if (mask != 0) {
    mask &= 0x8888888888888888ull;
    return offset + __builtin__ctzll(mask) >> 2;
  }
<snip>
```

It seems like this would eliminate an additional `fmov` and branch in the case there is an escape character. 

Admittedly I tried both versions of the code in `ruby/json` and at least on my M1 Macbook Air I really don't see much of a difference at least in the very coarse level benchmarking I've performed using `activitypub.json`. 

Is there a reason the `shrn` approach is inferior?

PR #2630, opened 2026-03-15
fix failing CI and rename misleading workflow name

jaja360:
I noticed the `Performance check on Ubuntu 20.04 CI (GCC 9)` workflow was not passing due to a cast warning and the use of `-Werror` in the workflow. This PR:

- Fix the casts in the code that were triggering the warning and crashing the workflow.
- Rename the misleading workflow names (they're called `Ubuntu 20.04 CI (GCC 9)`, but actually run on `Ubuntu 24.04 (GCC 13)`.

Type of change
- [ ] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests
- [x] Other (please describe): CI


lemire:
Merging.

PR #2662, opened 2026-04-05
Add guard for depth overflow in ondemand parser

uwezkhan:
Adds a check in skip_child() to avoid overflow in depth tracking.

In extreme cases, the depth counter can overflow when parsing
deeply nested input. This adds a guard before incrementing it
and returns an error if the limit is reached.

### DIFF
--- a/include/simdjson/generic/ondemand/json_iterator-inl.h
+++ b/include/simdjson/generic/ondemand/json_iterator-inl.h
@@ -158,6 +158,7 @@ simdjson_warn_unused simdjson_inline error_code json_iterator::skip_child(depth_
     switch (*return_current_and_advance()) {
       case '[': case '{':
         logger::log_start_value(*this, "skip");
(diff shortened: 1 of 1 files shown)


lemire:
Please demonstrate that an overflow is possible. I don't think it is.

uwezkhan:
You are right that this requires a very large and deeply nested input.

However the overflow is still reachable within the documented limits.
The depth counter is int32_t, and the maximum document size allows enough
characters to exceed INT32_MAX nesting levels .

In that case  _depth will wrap and the condition in skip_child() can
evaluate incorrectly causing the function to return early while still
inside the structure.

 I agree this is an extreme edge case. The change here is only
a small defensive guard before incrementing the counter, with no impact
on normal inputs.

lemire:
I am closing this. Please provide a reproducible test case.

PR #2665, opened 2026-04-07
Return CAPACITY instead of terminating on oversized max_depth in parser allocate APIs

jmestwa-coder:
### Description
Fixes a case where passing an oversized `max_depth` to parser allocation APIs could terminate the process instead of returning an error.

Previously, extreme values (e.g., `SIZE_MAX` and boundary variants) could reach `new[]` in a `noexcept` context, triggering `std::bad_array_new_length` and causing `std::terminate`.  
This change introduces a conservative bound on depth-related allocations and validates inputs before allocation, ensuring invalid values return `CAPACITY` consistently.

---

### Type of change
- [x] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests
- [ ] Other

---

### How to verify / test
- Added `parser_max_depth_tests` using public APIs.
- Verified behavior with boundary values:
  - `SIZE_MAX`
  - `SIZE_MAX / 2`
  - `SIZE_MAX / 8`
  - `SIZE_MAX / 16`
- Before fix: process terminates (`std::bad_array_new_length`)
- After fix: returns `CAPACITY` without termination
- Existing tests (`errortests`, `ondemand_error_tests`) pass unchanged


lemire:
I am going to fold this into a simpler PR. With credit to you.

Issue #2615, opened 2026-02-25
add feature macros


lemire:
Some users only rely on a small part of the simdjson library. To speed up compilation time, we could introduce feature macros whereas only some of the code gets compiled.

Raju1173:
A rough grouping of possible “feature units” based on the current structure :

- the two APIs :
-- ondemand
-- dom

- backends :
-- architecture specific implementation macros to explicitly disable specific architecture targets if a user knows their deployment environment.

We could perhaps introduce a central header file to keep the feature toggles centralized

lemire:
>  architecture specific implementation macros to explicitly disable specific architecture targets if a user knows their deployment environment.

This should be possible already.

Raju1173:
Thanks for the clarification, I’ve opened a PR that introduces feature flags to selectively disable the DOM and OnDemand APIs for now, as a first step towards more robust feature control...

Open for feedback.

lemire:
Might be a dead-end, see https://github.com/simdjson/simdjson/pull/2671

lemire:
This was explored in #2671. Putting the DOM, builder and On Demand parts behind feature macros turned out to be a dead end: the components are tightly coupled, and the approach would make the library harder to maintain without a clear compile-time benefit. You can already disable specific architecture backends with the existing `SIMDJSON_IMPLEMENTATION_*` macros. Closing for now.

PR #2670, opened 2026-04-09
build : added feature flags to selectively disable DOM and OnDemand APIs

Raju1173:
# Description

This PR introduces compile time feature flags to allow users to selectively enable or disable simdjson’s DOM and OnDemand APIs.
Related to #2615 

# Changes

Added two feature flags :

- SIMDJSON_FEATURE_DOM_API
- SIMDJSON_FEATURE_ONDEMAND_API

Both flags default to ON.

Added a compile-time check to enforce that at least one API remains enabled :
```
#if !SIMDJSON_FEATURE_DOM_API && !SIMDJSON_FEATURE_ONDEMAND_API
#error "simdjson: at least one of SIMDJSON_FEATURE_DOM_API or SIMDJSON_FEATURE_ONDEMAND_API must be enabled."
#endif
```
# Impact

Disabling DOM API:
- ~50% reduction in static library size (262K → 130K)
- ~35–40% reduction in build time (~8s → ~5s)

Disabling OnDemand API:
- minimal impact on binary size or build time

#

All valid feature combinations were tested with GCC 13.3.0 (WSL ubuntu)


lemire:
Looks like a dead-end, see https://github.com/simdjson/simdjson/pull/2671

PR #2671, opened 2026-04-09
build : added feature flags to selectively disable DOM, builder and OnDemand APIs

lemire:
Commits to make https://github.com/simdjson/simdjson/pull/2670 work.

Sadly, this proves useless. It does not affect the size of our binaries in a way that matters.

```

Test                      APIs                       lib time   simdjson.o   test time      test .o
---------------------------------------------------------------------------------------------------
dom_only                  DOM                          0.80 s      87.3 KB      0.72 s      67.6 KB
ondemand_only             DOM + ONDEMAND               0.58 s      87.3 KB      0.73 s      67.4 KB
dom_and_ondemand          DOM + ONDEMAND               0.57 s      87.3 KB      0.73 s      70.3 KB
builder_and_dom           DOM + BUILDER                0.57 s      87.3 KB      0.67 s      69.5 KB
builder_and_ondemand      DOM + ONDEMAND + BUILDER     0.57 s      87.3 KB      0.75 s      70.6 KB
builder_dom_and_ondemand  DOM + ONDEMAND + BUILDER     0.57 s      87.3 KB      0.77 s      76.5 KB
builder_only              DOM + BUILDER                0.66 s      87.3 KB      0.75 s      67.9 KB
```


Raju1173:
Thanks for investigating this further.

I hadn't considered that most users consume simdjson via the single header amalgamation and I was initially measuring the static library size (libsimdjson.a)...

Is there any path forward, such as generating feature specific single headers, or would that be too complex to maintain?

lemire:
> I hadn't considered that most users consume simdjson via the single header amalgamation and I was initially measuring the static library size (libsimdjson.a)...

I am not sure this matters. Whether one uses the single header or the regular version, one should end up with the same binary. I think.

Raju1173:
> I hadn't considered that most users consume simdjson via the single header amalgamation and I was initially measuring the static library size (libsimdjson.a)...

That was a miscommunication, the two sentences were supposed to be separate...

I tried the feature specific headers approach though, but the tight coupling of the internal components would make the library harder to maintain. So this does really seem like a dead end...

PR #2675, opened 2026-04-11
Fix integer overflow in allocation sizing to prevent OOB writes

metsw24-max:
This PR hardens allocation size computations in the parser to prevent integer overflow during capacity scaling and rounding operations.

The change ensures that all size calculations fail safely instead of silently wrapping on overflow.

-Introduced overflow-checked arithmetic for all allocation size computations
-Added guards before:
  - capacity scaling
  - padding addition
  - rounding operations
 
-Return `CAPACITY` error when overflow is detected (fail-closed behavior)

### DIFF
--- a/include/simdjson/dom/document-inl.h
+++ b/include/simdjson/dom/document-inl.h
@@ -10,6 +10,7 @@
 #include "simdjson/internal/jsonformatutils.h"
 
 #include <cstring>
+#include <limits>
 
 namespace simdjson {
 namespace dom {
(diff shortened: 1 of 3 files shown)


lemire:
@metsw24-max Thanks for the PR. I am not going to take this.

metsw24-max:
can i know reason for closing this pr?

lemire:
@metsw24-max The code is hard to read.

PR #2680, opened 2026-04-14
Fix undefined behavior in document_stream parsing

metsw24-max:
This fixes undefined behavior in stream parsing caused by passing potentially negative char values directly to `std::isspace`

All calls to std::isspace in the affected code paths have been updated to safely cast the input:

          std::isspace(static_cast<unsigned char>(...))

This follows the standard and recommended pattern for safe usage of C character classification functions.


lemire:
Fair. 

Let us run tests.

PR #2682, opened 2026-04-15
Prevent integer overflow in string buffer size computation

metsw24-max:
This patch fixes an integer overflow risk in the On-Demand parser allocation path.

The previous computation used to derive the string buffer size:

       5 * capacity / 3 + SIMDJSON_PADDING

could overflow due to the intermediate multiplication `5 * capacity` leading to wraparound and potentially undersized allocations This creates a memory safety risk under extreme input values


lemire:
This code is hard to read. We are not going to use this. 

Thanks for reporting the issue.

Issue #2684, opened 2026-04-16
On-Demand value::raw_json() after document::at_path_with_wildcard() returns a string_view that includes sibling fields and outer array delimiters


lifulong:
**Describe the bug**
When using the On-Demand API, simdjson::ondemand::document::at_path_with_wildcard() returns a std::vector<ondemand::value> of matches. For a match that is a nested JSON object (e.g. a field selected from an array element), calling raw_json() on that value can return a std::string_view whose suffix extends past the closing } of the intended object, including comma-separated sibling keys from the parent object and sometimes trailing ] from an enclosing array.

Typed access (e.g. get_object() + field getters) on the same match behaves correctly, so the semantic value is fine; only the raw JSON span from raw_json() is wrong.

**To Reproduce**

Expected behavior
For JSON:

[{"tag_meta":{"meta_code":2000211,"meta_value":""},"tag_value":""}]

and path "$[*].tag_meta", the single match should serialize as exactly:

{"meta_code":2000211,"meta_value":""}

Actual behavior

{"meta_code":2000211,"meta_value":""},"tag_value":""}]

raw_json() on the matched ondemand::value may return a longer prefix that starts with the correct object but ends after ,"tag_value":""}] (or similar), i.e. it leaks sibling and outer structure into the view.

Reproduce cpp code

```
// Repro: ondemand at_path_with_wildcard + raw_json() includes sibling fields.
// Build: c++ -std=c++17 -O2 -I.../include repro.cpp -L.../lib -lsimdjson -o repro

#include <simdjson.h>
#include <cstdlib>
#include <iostream>
#include <string_view>

int main() {
  using namespace simdjson;

  const padded_string json =
      R"([{"tag_meta":{"meta_code":2000211,"meta_value":""},"tag_value":""}])"_padded;
  constexpr std::string_view kPath = "$[*].tag_meta";
  constexpr std::string_view kExpected =
      R"({"meta_code":2000211,"meta_value":""})";

  ondemand::parser parser;
  ondemand::document doc = parser.iterate(json);

  auto wc = doc.at_path_with_wildcard(kPath);
  if (wc.error() || wc.value_unsafe().size() != 1) {
    return 1;
  }

  std::string_view raw;
  if (wc.value_unsafe()[0].raw_json().get(raw)) {
    return 1;
  }

  const bool ok =
      (raw == kExpected) && (raw.find("tag_value") == std::string_view::npos);
  if (!ok) {
    std::cerr << raw << '\n';
[...]

PR #2685, opened 2026-04-16
fix: ondeman raw_json boundaries for wildcard matches

hiteshmk05:
Short title (summary):

Description
- Fixed a bug in On-Demand wildcard extraction where `ondemand::value::raw_json` could return a `string_view` that extended past the matched object/array and included sibling or outer JSON content.
- Related issue: https://github.com/simdjson/simdjson/issues/2684

Type of change
- [x] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

How to verify / test
- Added a regression test in `ondemand_wildcard_tests.cpp` covering `$[*].tag_meta`.

### DIFF
--- a/include/simdjson/generic/ondemand/array-inl.h
+++ b/include/simdjson/generic/ondemand/array-inl.h
@@ -92,13 +92,20 @@ simdjson_warn_unused simdjson_warn_unused simdjson_inline error_code array::cons
 }
 
 simdjson_inline simdjson_result<std::string_view> array::raw_json() noexcept {
-  const uint8_t * starting_point{iter.peek_start()};
-  auto error = consume();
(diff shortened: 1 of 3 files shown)


lemire:
@hiteshmk05 I think that the bug is deeper.

hiteshmk05:
> @hiteshmk05 I think that the bug is deeper.

will take a more closer look this after work, will share my findings.

lemire:
It is a design issue. And it is my fault. I'll have a look because I don't think that the API works.

lemire:
See PR https://github.com/simdjson/simdjson/pull/2686

PR #2686, opened 2026-04-16
fix: ondemand for wildcard matches 

### LINKED ISSUE #2684: On-Demand value::raw_json() after document::at_path_with_wildcard() returns a string_view that includes sibling fields and outer array delimiters

lemire:
The design in https://github.com/simdjson/simdjson/pull/2533 by @hiteshmk05 cannot work. It is fatally flawed and led to bug https://github.com/simdjson/simdjson/issues/2684 which is not really a bug per se,  but just the result of our bad design.

It is not  @hiteshmk05's fault, it is my fault. I should not have merged it. You cannot capture `ondemand::value` in an `std::vector` to consume them later. They represent the value at the current time in the iteration and are meant to be consumed or discarded, not saved.

A proper way to fix the issue is to use the `for_each` construction where we pass a lambda and have the lambda consume the values.

Fixes https://github.com/simdjson/simdjson/issues/2684

### DIFF
--- a/doc/basics.md
+++ b/doc/basics.md
@@ -28,7 +28,7 @@ separate document](https://github.com/simdjson/simdjson/blob/master/doc/builder.
 - [UTF-8 validation (alone)](#utf-8-validation-alone)
 - [JSON Pointer](#json-pointer)
 - [JSONPath](#jsonpath)
-  * [Using `at_path_with_wildcard` for JSONPath Queries (On-Demand)](#using-at_path_with_wildcard-for-jsonpath-queries-on-demand)
+  * [Using `for_each_at_path_with_wildcard` for JSONPath Queries (On-Demand)](#using-for_each_at_path_with_wildcard-for-jsonpath-queries-on-demand)
     + [Example Usage](#example-usage)
 - [C++20 Ranges Support](#c20-ranges-support)
 - [Compile-Time JSONPath and JSON Pointer (C++26 Reflection)](#compile-time-jsonpath-and-json-pointer-c26-reflection)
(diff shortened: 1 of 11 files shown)


hiteshmk05:
> The design in https://github.com/simdjson/simdjson/pull/2533 by @hiteshmk05 cannot work. It is fatally flawed and led to bug https://github.com/simdjson/simdjson/issues/2684 which is not really a bug per se,  but just the result of our bad design.
> 
> It is not  @hiteshmk05's fault, it is my fault. I should not have merged it. You cannot capture `ondemand::value` in an `std::vector` to consume them later. They represent the value at the current time in the iteration and are meant to be consumed or discarded, not saved.
> 
> A proper way to fix the issue is to use the `for_each` construction where we pass a lambda and have the lambda consume the values.
> 
> Fixes https://github.com/simdjson/simdjson/issues/2684
> 

Makes sense. Thanks for the clarification.

PR #2687, opened 2026-04-17
Improve overflow-safe allocation sizing checks

metsw24-max:
This PR keeps the overflow hardening from #2675 but rewrites it to be easier to read and review.

What changed:

Refactored allocation-size checks into small named steps in:
`document-inl.h`
`dom_parser_implementation.h`
Kept fail-closed behavior: return `CAPACITY` on any overflow-risk path.
Added/kept the 32-bit regression guard in:
`errortests.cpp`


lemire:
Sorry no. This is messy and ugly code.

I don't want to receive such PRs in the future.

PR #2696, opened 2026-04-23
Add support for parsing NaN and Infinity as requested in #1540, #2414, and #2540

codeinred:
Summary: Add support for parsing NaN and Infinity as requested in #1540, #2414, and #2540

### Description

I added support for parsing NaN and Infinity when parsing doubles, or when using `get_double_in_string()`. Support for NaN and Infinity is gated behind `SIMDJSON_ENABLE_NAN_INF`. Tests were added for this feature.

`SIMDJSON_ENABLE_NAN_INF` is disabled by default, but may be enabled by passing `-DSIMDJSON_ENABLE_NAN_INF=ON` to CMake.

When `SIMDJSON_ENABLE_NAN_INF` is enabled, parsing is permissive:

- `nan` is matched (case-insensitive)
- `inf` is matched (case-insensitive).
- ~~`Infinity` is matched, as produced by python (requires exact match)~~
- `infinity` is matched (case-insensitive)

A `-` may appear before nan and infinity. This has no effect on the value of the NaN, but produces `-inf` when applied to infinity.

NaN and Infinity are only parsed if standard number parsing fails (eg, if the first character is not a digit in `parse_double`.

NaN and Infinity may also appear in strings, in which case they can be extracted with `parse_double_in_string`.

In both cases NaN and Infinity are only parsed if regular number parsing would otherwise fail. Any impact on performance should therefore be negligible.

In atom handling, if it fails to match a `null` atom, it will then check for a `nan` atom.

### Relevant issues

 #1540 and #2540. Also mentioned as a feature request in #2414.

### Type of change

- [ ] Bug fix
- [ ] Optimization
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

### How to verify / test

Additional tests were added for this feature. These are:

- `tests/dom/nan_inf_tests.cpp`
- `tests/ondemand/ondemand_nan_inf_tests.cpp`

A few tests which previously checked for failures when NaN/Infinity appeared were updated to validate that a NaN or Infinity value is correctly parsed (iff `SIMDJSON_ENABLE_NAN_INF` is enabled).

### Checklist before submitting

- [x] I linked the related issue (if applicable)

**Documentation:** I still need to update documentation. If you're happy with the design choices made by my PR, I can do so.

**Note:** To make the change smaller, I did _not_ update `singleheader/simdjson.h` or `singleheader/simdjson.cpp`. I can update these if needed.

### AI Disclosure

Claude was used to update CMake to add the `SIMDJSON_ENABLE_NAN_INF` feature flag. Claude was also used to write an initial version of the ondemand and nan/inf tests, which then underwent revision and review.

I requested that Claude adhere to simdjson conventions with regard to CMake and with regard to the tests, and based on my review I believe this to be the case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -254,6 +254,12 @@ if(SIMDJSON_MINUS_ZERO_AS_FLOAT)
   simdjson_add_props(target_compile_definitions PRIVATE SIMDJSON_MINUS_ZERO_AS_FLOAT=1)
 endif(SIMDJSON_MINUS_ZERO_AS_FLOAT)
 
+option(SIMDJSON_ENABLE_NAN_INF "Allow parsing of NaN and Infinity JSON values" OFF)
+if(SIMDJSON_ENABLE_NAN_INF)
+  message(STATUS "simdjson NaN and Infinity parsing is enabled.")
+  simdjson_add_props(target_compile_definitions PUBLIC SIMDJSON_ENABLE_NAN_INF=1)
+endif()
+
 # GCC and Clang have horrendous Debug builds when using SIMD.
 # A common fix is to use '-Og' instead.
 # bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412
--- a/include/simdjson/common_defs.h
+++ b/include/simdjson/common_defs.h
@@ -30,6 +30,10 @@ double from_chars(const char *first, const char* end) noexcept;
 #endif
 #endif
 
+#ifndef SIMDJSON_ENABLE_NAN_INF
+#define SIMDJSON_ENABLE_NAN_INF 0
+#endif
+
 } // namespace simdjson
 
 #if defined(__GNUC__)
--- a/tests/ondemand/ondemand_nan_inf_tests.cpp
+++ b/tests/ondemand/ondemand_nan_inf_tests.cpp
@@ -1,0 +1,352 @@
+#include "simdjson.h"
+#include "test_ondemand.h"
+#include <array>
+#include <cmath>
+#include <limits>
+
+using namespace simdjson;
+
+namespace nan_inf_tests {
+#if SIMDJSON_ENABLE_NAN_INF
+
+bool parse_nan() {
+  TEST_START();
+  for (auto json_str : {"NaN", "nan", "-nan", "-NAN"}) {
+    ondemand::parser parser;
+    padded_string json(json_str, strlen(json_str));
+    ondemand::document doc;
+    ASSERT_SUCCESS(parser.iterate(json).get(doc));
+    double value;
+    ASSERT_SUCCESS(doc.get_double().get(value));
+    ASSERT_TRUE(std::isnan(value));
+  }
+  TEST_SUCCEED();
+}
+
+bool parse_infinity() {
+  TEST_START();
+  for (auto json_str : {"Infinity", "inf", "INF", "Inf"}) {
+    ondemand::parser parser;
+    padded_string json(json_str, strlen(json_str));
+    ondemand::document doc;
+    ASSERT_SUCCESS(parser.iterate(json).get(doc));
+    double value;
+    ASSERT_SUCCESS(doc.get_double().get(value));
+    ASSERT_TRUE(std::isinf(value));
+    ASSERT_TRUE(value > 0);
+  }
+  TEST_SUCCEED();
+}
+
+bool parse_negative_infinity() {
+  TEST_START();
+  for (auto json_str : {"-Infinity", "-inf", "-INF", "-Inf"}) {
+    ondemand::parser parser;
+    padded_string json(json_str, strlen(json_str));
+    ondemand::document doc;
+    ASSERT_SUCCESS(parser.iterate(json).get(doc));
+    double value;
+    ASSERT_SUCCESS(doc.get_double().get(value));
+    ASSERT_TRUE(std::isinf(value));
+    ASSERT_TRUE(value < 0);
+  }
+  TEST_SUCCEED();
+}
+
+bool nan_in_array() {
+  TEST_START();
+  ondemand::parser parser;
+  auto json = R"([5, NaN, NAN, nan, -nan, -NAN, -NaN, 1.25])"_padded;
+  ondemand::document doc;
+  ASSERT_SUCCESS(parser.iterate(json).get(doc));
+  ondemand::array arr;
+  ASSERT_SUCCESS(doc.get_array().get(arr));
+
+  double nan = std::numeric_limits<double>::quiet_NaN();
+  std::array<double, 8> expected_values{5, nan, nan, nan, nan, nan, nan, 1.25};
+
+  size_t index = 0;
+  for (auto val : arr) {
+    if (index == expected_values.size()) {
+      TEST_FAIL("Array contained more values than expected");
+    }
+
+    double parsed;
+    ASSERT_SUCCESS(val.get_double().get(parsed));
+    double expected = expected_values[index];
+    if (std::isnan(expected)) {
+      ASSERT_TRUE(std::isnan(parsed))
+    } else {
+      ASSERT_EQUAL(parsed, expected_values[index]);
+    }
+
+    index++;
+  }
+
+  ASSERT_EQUAL(index, expected_values.size());
+  TEST_SUCCEED();
+}
+
+bool infinity_in_array() {
+  TEST_START();
+  ondemand::parser parser;
+  auto json = R"([1,
+      infinity,
+      INFINITY,
+      Infinity,
+      inf,
+      Inf,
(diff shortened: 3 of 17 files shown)


CMakeLists.txt
```diff
@@ -254,6 +254,12 @@ if(SIMDJSON_MINUS_ZERO_AS_FLOAT)
   simdjson_add_props(target_compile_definitions PRIVATE SIMDJSON_MINUS_ZERO_AS_FLOAT=1)
 endif(SIMDJSON_MINUS_ZERO_AS_FLOAT)
 
+option(SIMDJSON_ENABLE_NAN_INF "Allow parsing of NaN and Infinity JSON values" ON)
+if(NOT SIMDJSON_ENABLE_NAN_INF)
```
lemire: That's likely backward. You want to set the macro to 1 when it is enabled.

CMakeLists.txt
```diff
@@ -254,6 +254,12 @@ if(SIMDJSON_MINUS_ZERO_AS_FLOAT)
   simdjson_add_props(target_compile_definitions PRIVATE SIMDJSON_MINUS_ZERO_AS_FLOAT=1)
 endif(SIMDJSON_MINUS_ZERO_AS_FLOAT)
 
+option(SIMDJSON_ENABLE_NAN_INF "Allow parsing of NaN and Infinity JSON values" ON)
```
lemire:
```suggestion
option(SIMDJSON_ENABLE_NAN_INF "Allow parsing of NaN and Infinity JSON values" OFF)
```

include/simdjson/common_defs.h
```diff
@@ -30,6 +30,10 @@ double from_chars(const char *first, const char* end) noexcept;
 #endif
 #endif
 
+#ifndef SIMDJSON_ENABLE_NAN_INF
+#define SIMDJSON_ENABLE_NAN_INF 1
```
lemire:
```suggestion
#define SIMDJSON_ENABLE_NAN_INF 0
```

lemire: We are NOT enabling this by default.

This does not abide by the JSON spec. Proper JSON does not generate NaN or Inf as per the spec.

tests/ondemand/ondemand_nan_inf_tests.cpp
```diff
@@ -0,0 +1,277 @@
...
+      double value;
+      ASSERT_SUCCESS(doc.get_double().get(value));
+      ASSERT_TRUE(std::isnan(value));
+    }
+    TEST_SUCCEED();
+  }
+
+  bool parse_infinity() {
+    TEST_START();
+    for (auto json_str : {"Infinity", "inf", "INF", "Inf"}) {
```
lemire:
```suggestion
    for (auto json_str : {"infinity", "Infinity", "inf", "INF", "Inf"}) {
```

tests/ondemand/ondemand_nan_inf_tests.cpp
```diff
@@ -0,0 +1,277 @@
...
+      ASSERT_SUCCESS(doc.get_double().get(value));
+      ASSERT_TRUE(std::isinf(value));
+      ASSERT_TRUE(value > 0);
+    }
+    TEST_SUCCEED();
+  }
+
+  bool parse_negative_infinity() {
+    TEST_START();
+    for (auto json_str : {"-Infinity", "-inf", "-INF", "-Inf"}) {
```
lemire:
```suggestion
    for (auto json_str : {"-infinity", "-Infinity", "-inf", "-INF", "-Inf"}) {
```

CMakeLists.txt
lemire: We are not enabling support for broken JSON by default.

The default is proper JSON. Improper JSON should be an option, not a requirement.

tests/ondemand/ondemand_nan_inf_tests.cpp
codeinred: Just to clarify, would you also like me to handle case-insensitive checks for "Infinity"? With the current implementation, only the short version is case-insensitive

include/simdjson/common_defs.h
codeinred: I will fix the CMake so that it is not enabled by default

tests/ondemand/ondemand_nan_inf_tests.cpp
lemire: There is no specification (by definition, you are dealing with broken JSON), but what  happens if you run this:

```cpp
#include <iostream>
#include <iostream>
#include <string>
#include <stdexcept>

double parse_double(const std::string& s) {
    try {
        size_t pos = 0;
        double value = std::stod(s, &pos);

        // Optional: ensure the whole string was consumed
        if (pos != s.length()) {
            throw std::invalid_argument("Extra characters after number");
        }
        return value;
    } catch (...) {
        throw std::invalid_argument("Invalid double: " + s);
    }
}

int main() {
    std::string tests[] = {"infinity", "inFinity"};

    for (const auto& t : tests) {
        std::cout << t << " -> " << parse_double(t) << '\n';
    }
}
```

So many string to double parsers will use a case insensitive approach. Now it  might be fine not to do this, but it should be documented.

lemire:
@codeinred Just so you understand why it needs to be disabled by default:

Capture d’écran, le 2026-04-24 à 12 18 16


Anyone producing JSON with Infinity or some such is producing broken JSON that will not be parseable in, say, JavaScript. It defeats the purpose of JSON as a universal/portable format. Once we start breaking the spec, all sorts of nasty issues come up. So I strongly discourage it.

I am fine with disabled-by-default features though.

tests/ondemand/ondemand_unknown_tests.cpp
```diff
@@ -31,8 +33,13 @@ namespace unknown_tests {
         auto val = object["key"];
         ASSERT_SUCCESS( val.type().get(type) );
         ASSERT_EQUAL( type, simdjson::ondemand::json_type::unknown );
-        double num;
+        double num = 0.0;
+#if SIMDJSON_ENABLE_NAN_INF
+        ASSERT_EQUAL( val.get(num), simdjson::error_code::SUCCESS );
```
lemire:
```suggestion
        ASSERT_SUCCESS( val.get(num));
```

tests/ondemand/ondemand_unknown_tests.cpp
lemire: So here the construction does not work. ASSERT_TRUE is a macro, and the PR passes a macro inside the macro. It gets too complicated.

What you want to do is start with the condition


```cpp
#if SIMDJSON_ENABLE_NAN_INF
ASSERT_TRUE(te...
#else  // SIMDJSON_ENABLE_NAN_INF
ASSERT_TRUE(te...
#endif  // SIMDJSON_ENABLE_NAN_INF
```

lemire:
@codeinred If you look at my comments, the Visual Studio errors ought to be easy to fix. 


Otherwise, this looks like good work to me.

lemire:
@codeinred Note that I don't care whether `infinity` is supported or not, your PR description is clear, but I encourage you to write it out in the code.

codeinred:
Hi @lemire! I appreciate all of your feedback. I have updated my PR with the following changes:

- I updated the CMake so that extended parser support for NaN/Inf is now disabled by default, as requested.
- I updated the parser to support case-insensitive parsing of `infinity` (8 characters) in addition to the shorter version. 
- I updated the tests to check for case-insensitive infinity
- Per your recommendation, I fixed the use of `#if SIMDJSON_ENABLE_NAN_INF` inside a macro argument, and I've confirmed that the library compiles and tests pass with MSVC (I tested Visual Studio 2022, but I expect it to work on earlier versions as well).

  This last suggestion was particularly helpful, and saved me some time debugging - thank you!

I've cleaned up the commit history as well, and I attempted to update a few CI/CD workflows in order to also test the library with support for NaN/Infinity enabled. This is the top commit.

image

This was my attempt to save you the work of updating CI, but please feel free to discard that commit if you do not like the changes. Please let me know if you have additional feedback!

include/simdjson/common_defs.h
codeinred: [Updated PR to disable by default]

tests/ondemand/ondemand_nan_inf_tests.cpp
codeinred: [Updated PR to support case-insensitive matching for both 'inf' and 'infinity']

tests/ondemand/ondemand_nan_inf_tests.cpp
codeinred: [Updated PR to support case-insensitive matching for both 'inf' and 'infinity']

lemire:
Will be part of our next major release.

PR #2697, opened 2026-04-25
Add support for custom allocators

vogelsgesang:
**Description**: Support for custom allocators (addresses #1017).

I also added support for [allocate_at_least](https://en.cppreference.com/cpp/memory/allocator/allocate_at_least)-style overallocations.
(The use case I would like to integrate simdjson into could benefit from it)

**Type of change**: New Feature
**How to verify / test**: New test cases added as part of the PR

**AI Disclaimer**:

**Please don't review the code line-by-line yet** -- I would first like
high-level alignment on the API shape and the rationale. I am happy
to redo the implementation once we agree on the direction.

This PR is intended as the discussion vehicle for adding custom-allocator
support; per CONTRIBUTING.md I would normally open an issue first, but
the design is concrete enough that a PR seemed clearer and drafting the
code change was not too much effort thanks to Claude & Cursor.

**Open design questions**

- Is the general approach (custom abstract base class + RAII buffer
  wrapper) acceptable, or would you prefer a different shape entirely
  (templated, `std::pmr`-only, hooks, ...)?
- Are we OK with threading the allocator through so many API layers
  (parser → document → `dom_parser_implementation` → per-kernel
  subclasses)? It is mechanical but still a bit of churn.
- I had to drop `noexcept` from a number of public methods because a
  user allocator may throw. Three options:
  1. Drop `noexcept` unconditionally (current PR).
  2. Drop `noexcept` only when `SIMDJSON_EXCEPTIONS` is set.
  3. Gate behind a new `SIMDJSON_SUPPORT_THROWING_ALLOCATOR` macro,
     defaulting to off.
- Do we actually care about *strong* exception safety? If not, several
  reallocation sites can be simplified by skipping the
  allocate-into-local-then-move-assign dance.

----

**High-level Design**

We use a virtual base class `simdjson::allocator`. This class provides virtual
`allocate` and `deallocate` functions.

A `simdjson::allocator&` is threaded through the various classes / interfaces,
so we have it available at all allocation sites.

Users who already have a `std::pmr::memory_resource` or a
`std::allocator<T>`-conforming type can plug it in via
`simdjson::allocator_adapter<Alloc>`.

**Why not `std::pmr` or `std::allocator` directly?**

PR #1467 attempted to use `std::allocator` and was closed because of
three issues:

1. `std::allocator::allocate` throws `std::bad_alloc` -- incompatible
   with simdjson's error-code model.
2. Mixing `operator new[]` / `delete[]` with allocator-allocated memory
   produced alloc-dealloc-mismatch under sanitizers.
3. `std::unique_ptr<T[]>` does not natively support stateful allocators;
   per-allocation custom deleters get unwieldy.

I tried to address all 3 issues in this design.

**The allocator interface**

```cpp
struct allocation_result {
  void* ptr;
  size_t size;  // actual bytes allocated, >= requested
};

class allocator {
public:
  virtual ~allocator() = default;

  /**
   * Allocate AT LEAST `size` bytes. Return `{nullptr, 0}` on failure, OR throw.
   * The returned `size` may exceed the request; pass it back to `deallocate`
   * unchanged. Modelled on std::allocator::allocate_at_least (C++23).
   */
  virtual allocation_result allocate(size_t size) = 0;

  /**
   * Deallocate memory previously returned by `allocate()`.
   * `size` must equal `allocation_result::size` from the corresponding
   * `allocate()` call (sized-deallocation style, like C++14
   * `operator delete(void*, size_t)`).
   */
  virtual void deallocate(void* ptr, size_t size) noexcept = 0;
};
```

Notes:

- `allocate` *may* throw, but the default never does, so existing
  behaviour is preserved.
- The default allocator is a process-wide singleton, wrapping
  `new[] (std::nothrow)`. Parsers store a single `allocator*`; users
  who don't opt in pay no extra indirection on the hot path beyond
  one virtual call per (rare) allocation.
- `allocation_result::size` returns the *actual* bytes allocated. This
  is propagated into the parser's `_capacity` / `_max_depth` /
  `loaded_bytes.capacity()` etc., so when the allocator over-provisions
  (jemalloc size classes, huge-page rounding, slab allocators) the next
  slightly-larger document can often be parsed without reallocating at
  all. There are tests covering this for both `dom::parser` and the
  builder.

**RAII wrapper**

`internal::allocated_buffer<T>` replaces the `std::unique_ptr<T[]>` buffers.
It stores the byte size and the allocator pointer that produced the memory,
so sized-deallocation works correctly even after move-assigning between parsers
that use different allocators.

**Exception safety**

Reallocation sites allocate into a local `allocated_buffer` first and
only move-assign into the member on success. A failed allocation
(returned `nullptr` or thrown) leaves the parser in its previous usable
state -- strong exception safety. simdjson itself never catches
allocator exceptions; they propagate to the caller unchanged.

**Adapter for standard `Allocator` types**

For users that already have a type modelling the C++ `Allocator` named
requirement (`std::allocator<T>`, `std::pmr::polymorphic_allocator<T>`,
or a third-party pool's allocator), `simdjson::allocator_adapter<Alloc>`
wraps it as a `simdjson::allocator`. When `__cpp_lib_allocate_at_least`
is available it forwards to `std::allocator_traits::allocate_at_least`
so over-allocation surfaces through to simdjson's capacity tracking.

```cpp
simdjson::allocator_adapter<std::pmr::polymorphic_allocator<uint8_t>> a{
    std::pmr::polymorphic_allocator<uint8_t>{&my_resource}};
simdjson::ondemand::parser parser{a};
```

**Scope**

Applies to `dom::parser`, `dom::document`, `ondemand::parser`,
`dom_parser_implementation` (+ kernel subclasses), and
`builder::string_builder`.

Deliberately *not* applied to:

- `padded_string` / `padded_string_builder` -- input role symmetrical
  with `padded_string_view` (caller brings the buffer).
- `ondemand::parser::get_parser()` -- swappable allocator on a static
  thread-local would make lifetime reasoning brittle.

**Performance**

The default-allocator path is unchanged: same `new[] (std::nothrow)`,
no extra heap traffic, one extra `allocator*` member on each parser.
No measurable regression is expected; happy to run a before/after
parsing benchmark if you want hard numbers before merging.

### DIFF
--- a/doc/performance.md
+++ b/doc/performance.md
@@ -9,6 +9,7 @@ testing and get the best performance.
 * [NDEBUG macro](#ndebug-macro)
 * [Reusing the parser for maximum efficiency](#reusing-the-parser-for-maximum-efficiency)
 * [Reusing string buffers](#reusing-string-buffers)
+* [Custom allocators](#custom-allocators)
 * [Server Loops: Long-Running Processes and Memory Capacity](#server-loops-long-running-processes-and-memory-capacity)
 * [Large files and huge page support](#large-files-and-huge-page-support)
 * [Number parsing](#number-parsing)
@@ -81,6 +82,106 @@ or simply
 ```cpp
  auto doc = parser.iterate(json_str, length, capacity);
 ```
+
+
+Custom allocators
+-----------------
+
+By default, simdjson uses `new[]` / `delete[]` for its internal buffers. If you
+want to use a custom allocator instead, you can do so by subclassing
+`simdjson::allocator`:
+
+```cpp
+class my_allocator : public simdjson::allocator {
+public:
+  simdjson::allocation_result allocate(size_t size) override {
+    // Return {nullptr, 0} on failure, OR throw (e.g. std::bad_alloc). Either is fine.
+    void* ptr = ::operator new(size, std::nothrow);
+    return {ptr, ptr ? size : 0};
+  }
+  void deallocate(void* ptr, size_t /*size*/) noexcept override {
+    ::operator delete(ptr);
+  }
+};
+```
+
+Then pass an instance to the parser or builder:
+
+```cpp
+my_allocator alloc;
+
+// DOM parser
+simdjson::dom::parser dom_parser(alloc);
+auto e1 = dom_parser.parse(json);
+
+// On-Demand parser
+simdjson::ondemand::parser od_parser(alloc);
+auto e2 = od_parser.iterate(json);
+
+// String builder
+simdjson::builder::string_builder sb(alloc);
+sb.append("hello");
+```
+
+`padded_string` and `padded_string_builder` do not support custom allocators.
+If you want to allocate a padded string with a custom allocator, you can do
+so in your own code, and then pass it into the simdjson library as a
+`padded_string_view`.
+
+Key points:
+
+- The allocator must outlive every parser, document, and builder that holds a
+  reference to it.
+- `allocate()` may either return `{nullptr, 0}` on failure (mapped to `MEMALLOC`
+  or `is_valid = false` on the builder) or throw. Exceptions are not caught by
+  simdjson: they propagate out of whichever call triggered the allocation.
+- `allocate()` may over-allocate: the returned `allocation_result::size` is the
+  true number of bytes handed out and may exceed the requested `size`
+  (modelled on C++23 `allocate_at_least`). simdjson may use the full returned
+  range.
+- `deallocate()` is called with the same byte count that the corresponding
+  `allocate()` returned in `allocation_result::size`, not necessarily the
+  originally requested size. Allocators that round up to buckets or slabs and
+  hand back a larger `size` receive that same larger value back on
+  `deallocate()`.
+- Returned memory must be aligned to at least `alignof(std::max_align_t)`. This
+  matches the guarantee of `malloc`, `::operator new`, and all mainstream pool
+  allocators.
+- The thread-local parser returned by `simdjson::ondemand::parser::get_parser()`
+  always uses the default allocator. Use your own parser instance if you need
+  a custom allocator.
+
+### Adapting a C++ standard allocator
+
+If you already have a type that models the C++ `Allocator` named requirement
+(e.g. `std::allocator<T>`, `std::pmr::polymorphic_allocator<T>`, or a
+third-party pool's allocator), you can use it directly via
+`simdjson::allocator_adapter<Alloc>` instead of subclassing
+`simdjson::allocator` yourself:
+
+```cpp
+#include <simdjson/allocator_adapter.h>
+
+// std::allocator
+simdjson::allocator_adapter<std::allocator<uint8_t>> a1;
+simdjson::dom::parser p1(a1);
+
+// std::pmr
+std::pmr::monotonic_buffer_resource mbr;
+simdjson::allocator_adapter<std::pmr::polymorphic_allocator<uint8_t>> a2(
+    std::pmr::polymorphic_allocator<uint8_t>{&mbr});
+simdjson::dom::parser p2(a2);
+```
+
+The adapter rebinds the wrapped allocator to `uint8_t`, forwards
+`allocate` / `deallocate`, and uses `allocate_at_least` when available
+(C++23).
+
+Alignment caveat: Your allocator must return memory that is aligned to at
+least `alignof(std::max_align_t)`. All standard-library-provided resources
+and every mainstream pool allocator already do so, but for custom allocators,
+you need to ensure that your allocator returns memory that is aligned to at
+least `alignof(std::max_align_t)`.
 
 
 Server Loops: Long-Running Processes and Memory Capacity
--- a/include/simdjson/dom/document-inl.h
+++ b/include/simdjson/dom/document-inl.h
@@ -9,6 +9,7 @@
 #include "simdjson/internal/tape_ref-inl.h"
 #include "simdjson/internal/jsonformatutils.h"
 
+#include <algorithm>
 #include <cstring>
 
 namespace simdjson {
@@ -26,7 +27,7 @@ inline size_t document::capacity() const noexcept {
 }
 
 simdjson_warn_unused
-inline error_code document::allocate(size_t capacity) noexcept {
+inline error_code document::allocate(size_t capacity) {
   if (capacity == 0) {
     string_buf.reset();
     tape.reset();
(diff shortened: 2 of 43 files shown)


lemire:
Sorry for the delay.

1. I don't think we want to break the `noexcept`. It would be ok to do `noexcept(myallocator.except)` with a default allocator that is `noexcept`. The general rule should be to avoid breaking our public interface.
2. I am not too worried about performance and we will verify this anyhow.

vogelsgesang:
> Sorry for the delay.

No worries, thanks for the feedback!

> I don't think we want to break the noexcept. [...] The general rule should be to avoid breaking our public interface.

Sounds good.

Just so I understand the guidance: Is simdjson aiming for source-code compatibility or ABI compatibility? Afaict, source-code compatibility would be retained even if we remove the `noexcept`? (At least for most "just call the functions" usages)

>  It would be ok to do `noexcept(myallocator.except)` with a default allocator that is `noexcept`

That would only work if we would pass in the allocator type as a template parameter - that's not what this commit is doing, though. This commit doesn't pass the allocator as a template parameter but as a type-erased runtime parameter.

With (1) being a no-go, I see two remaining options

1. ~Drop noexcept unconditionally (current PR).~
2. Drop noexcept only when SIMDJSON_EXCEPTIONS is set.
3. Gate behind a new SIMDJSON_SUPPORT_THROWING_ALLOCATOR macro,
defaulting to off.

For my use case, both would be fine. I would go with (3) since it means less code churn.

lemire:
@vogelsgesang Can't we require the allocator to be noexcept ?

```cpp
T* allocate(std::size_t n) noexcept {
        try {
 // blabla
        } catch (...) {
            return nullptr;
        }
 }
```

vogelsgesang:
it's not perfect for my particular downstream use case, but works for me, too.

In my use case, the exceptions thrown by the allocator have additional payload information.

More details (feel free to skip)
We have two memory limits: one process-wide limit and a per-request limit.
We over-commit memory, i.e. we allow more requests to be in the system at the same time than we could serve if every request actually uses its full memory budget. We bet on the fact that most requests are simple and don't actually use the memory.
Our allocator instances track how much memory was already allocated per query and then throw the corresponding "per-request limit exceeded" or "global limit exceeded". Depending on which limit we reach, this counts differently against our SLA: If we hit the per-request limit, it's the users fault since he gave us a too complex query. This doesn't count against our SLA. On the other hand, if the query failed due to us over-comitting memory, it does count against our SLA, since this is an optimization done by us under the covers, outside the control of the user.

I see how this might be an edge case, and for the majority of users of simdjson, keeping the allocators noexcept might indeed be the better choice.

If we keep `allocate` as `noexcept` as you propose, I would simply do the following in my downstream code:

```cpp
threadlocal std::exception_ptr threadlocal_exception_ptr;

// My custom allocator, implemented in my downstream code
struct tracking_allocator : allocator {
  // ... <other functions> ...
  simdjson::allocation_result allocate(size_t size) override {
    try {
      return backing_allocator->allocate(size_t size);
    } catch(...) {
      threadlocal_exception_ptr = std::current_exception();
      return nullptr;
    }
  }
  // ... <other functions> ...
};

auto someCaller() {
  tracking_allocator alloc(1);
  simdjson::dom::parser parser(alloc);
  simdjson::dom::element elem;
  auto result = parser.parse(jsonStr, jsonStr.size()).get();
  if (result == simdjson::MEMALLOC) {
    std::rethrow_exception(threadlocal_exception_ptr);
  }
  // ...<more code>...
}
 ```

**Comment by lemire** [comment]
@vogelsgesang We can also do the macro thing. (We have lots of little macros for specific use cases.)

But let me go to something I feel is important. The way the simdjson lib is built, you can effectively avoid allocations. That's the point of the parser, we have it have a capacity and then from there, it can basically be reused. That's really how we hope people use it on long running services.

So I am a little worried that you might be doing lots of allocation and de-allocation. Of course, you do what you want... but I just want to make sure that there is no confusion as to what we expect.

**Comment by lemire** [comment]
@vogelsgesang So, here for example...

```cpp
auto someCaller() {
  tracking_allocator alloc(1);
  simdjson::dom::parser parser(alloc);
  simdjson::dom::element elem;
  auto result = parser.parse(jsonStr, jsonStr.size()).get();
  if (result == simdjson::MEMALLOC) {
    std::rethrow_exception(threadlocal_exception_ptr);
  }
...
```

We prefer that users do not allocate a parser each time they parse a document. The reason we have distinct type `parser` is precisely to control and limit allocations.

So let us say that each thread only ever parses one JSON document at any one time, then what we would suggest is a thread local parser.

And you can even put a limit on it (so that it throws or return an error if the input is too large). So you can create a parser, say that the parser can't exceed that much capacity and then reuse the parser.

Your memory usage will then be flat as far as the parser is concerned. If your purpose is to have bounded capacity, there is no need for a custom allocator at all. The functionality is built in!!!

vogelsgesang:
> But let me go to something I feel is important. The way the simdjson lib is built, you can effectively avoid allocations. That's the point of the parser, we have it have a capacity and then from there, it can basically be reused. That's really how we hope people use it on long running services.

Understood - Thank you for pointing out the potential performance pitfalls and that I should reuse the allocations!

**Long story short: Don't worry, (I think?) I know what I am doing**

My example regarding `auto someCaller()` was oversimplified and only focused on the the aspect of "how could I propagate the exception payload via a threadlocal" while ignoring the reuse of allocations. In practice, I will be reusing the parser. 

(Also, I will be using an `ondemand::parser` instead of a `dom::parser`. In fact, I would even prefer a parser which pre-parses the document into a "tape" similar to the DOM parser, but into a _single_ instead of _two_ tapes as the DOM parser currently does - but I am digressing...) 

**And the long version:** I am planning to integrate simdjson into a database system. This system will receive queries like

```
SELECT
  json_value(my_table.details, '$.stacktrace[0].symbol_name') AS crash_symbol,
  COUNT(*)
FROM log_events
WHERE json_value(my_table.metadata, '$.event.type') = 'crash'
GROUP BY crash_symbol
```

Note how:
1. I have an input table `log_events`. Let's assume it contains 1 million records
2. I have 2 `json_value` calls accessing JSON documents from 2 different columns (`metadata` and `details`)

-> I will be processing 2 million JSON documents for this query.

When starting to process this query, I would set up two `simdjson::ondemand::parser` instances in the thread-local storage of each thread, one for each of the `json_value` call sites.
For each new tuple from the `log_events`, I would reuse those two parsers.
After the query is done, I would tear down the parser instances.
I think the setup / teardown cost of two parsers per thread is acceptable for processing 2 million tuples.

> And you can even put a limit on it (so that it throws or return an error if the input is too large). So you can create a parser, say that the parser can't exceed that much capacity and then reuse the parser.

Afaict, I can only set a static limit via this mechanism and there is no callback mechanism "I am about to allocate X MB - is that ok right now?".

In my case, whether an allocation should be successful also depends on what the other parts of the query are doing. I want to limit the overall usage for the complete query, not only of each individual component. JSON-parsing is just one potential user of memory.

E.g., if the hash table for the `GROUP BY crash_symbol` is already taking up 99MB, then the simdjson parser would have only 1MB of the overall query budget of 100MB left. However, if the hash table happens to only take 5 MB, the simdjson parser would be allowed to use up to 95MB. 

The easiest way I could find to make such memory budgeting work across a wide set of data structures (tuple buffers, hash tables, regular expression engines, ...) was to pass each of those component the same (query-local) allocator. As soon as the memory allocated via that allocator goes over the budget, I throw an exception and thereby fail the query

lemire:
@vogelsgesang

We could also keep the style of the current PR and remove the `noexcept` *if* we can establish that it does not affect the performance. That would require a major release, of course.




> Afaict, I can only set a static limit via this mechanism and there is no callback mechanism "I am about to allocate X MB - is that ok right now?".

If the parser has max capacity X and you try to parse something larger, it will fail right away, with some kind of memory error. It is designed so that you can forbid the parser from allocating (at all).

vogelsgesang:
> We could also keep the style of the current PR and remove the noexcept if we can establish that it does not affect the performance. That would require a major release, of course.

The longer I think about it, the more I realize that removing the `noexcept` is actually the least important part of this change. I think I would just keep the `noexcept`.

From the earlier options

> 1. Drop noexcept unconditionally (current PR).
> 2. Drop noexcept only when SIMDJSON_EXCEPTIONS is set.
> 3. Gate behind a new SIMDJSON_SUPPORT_THROWING_ALLOCATOR macro, defaulting to off.

(2) and (3) also aren't nice, because throwing allocators would be a rarely used feature, and new code would probably add new unconditional `noexcept`s, which has a risk of inadvertently breaking the `SIMDJSON_SUPPORT_THROWING_ALLOCATOR` config. The [threadlocal approach](https://github.com/simdjson/simdjson/pull/2697#issuecomment-4362978285) is good enough for me.

I will update the PR soon(-ish) to keep the `noexcept` in place

include/simdjson/dom/document-inl.h
```diff
@@ -26,7 +27,7 @@ inline size_t document::capacity() const noexcept {
 }
 
 simdjson_warn_unused
-inline error_code document::allocate(size_t capacity) noexcept {
+inline error_code document::allocate(size_t capacity) {
```
hungptit, line 30: We currently return the error code and throw an exception as well. I'm not entirely clear on the reasoning here—would you mind clarifying the intended behavior?

doc/performance.md
```diff
@@ -9,6 +9,7 @@ testing and get the best performance.
 * [NDEBUG macro](#ndebug-macro)
 * [Reusing the parser for maximum efficiency](#reusing-the-parser-for-maximum-efficiency)
 * [Reusing string buffers](#reusing-string-buffers)
+* [Custom allocators](#custom-allocators)
```
hungptit, line 12: Thank you very much, @vogelsgesang, for leading this effort! I would also benefit greatly from this functionality in one of my projects.

PR #2700, opened 2026-04-25
Prevent overflow in parser allocation size calculations

uwezkhan:
This updates the allocation size calculations in document::allocate and
ondemand::parser::allocate to use overflow-safe arithmetic.

The previous code performed multiplication before division when computing
string buffer capacity, which could overflow size_t on some edge-case inputs
before the final bounds check.

This change moves the calculation to a wider intermediate type and adds an
explicit bounds check before casting back.

Normal allocation behavior is unchanged for valid inputs.


lemire:
Please stop spamming our project with garbage PR.

PR #2702, opened 2026-04-29
Prevent overflow in large-capacity parser allocations

uwezkhan:
This fixes overflow issues in parser capacity calculations when handling
very large capacity values near the supported maximum size.

A few allocation paths were doing rounding/multiplication on size_t values
before the final bounds checks, which could overflow on 32-bit builds and
lead to incorrect buffer sizing.

The fix adds precise overflow-safe guards and performs intermediate
capacity calculations in 64-bit arithmetic where needed.


uwezkhan:
Thanks for the earlier feedback — I took a more careful pass before opening this one.

This PR is limited to a specific allocation-arithmetic issue in the parser capacity calculations. The affected code can overflow during rounding/growth calculations near the maximum supported capacity on 32-bit builds, which may result in undersized allocations before parsing begins.

I kept the change narrowly scoped to the relevant arithmetic/rounding paths and avoided unrelated hardening/refactors.

Happy to adjust the approach if you’d prefer the checks structured differently.

lemire:
User @uwezkhan has been reported for spamming with low quality PRs.

This kind of behaviour will not be accepted.

PR #2704, opened 2026-04-30
Fix from(std::string) input lifetime (#2703)

### LINKED ISSUE #2703: from(std::string) does not keep its input alive
**Describe the bug**

  `simdjson::from(std::string)` can return an `auto_parser` whose on-demand document still refers to the buffer owned by the input string. If the input is a temporary or moved-from string, the actual conversion may happen after that
  buffer is no longer valid.

  This can lead to a use-after-free when the returned object is converted into the target type. The same issue applies to the overload that takes an external `ondemand::parser`.

  **To Reproduce**

  A reduced test is to pass a temporary `std::string` to `from(...)`, delay materialization, churn heap memory, and then convert the returned object. With sanitizers enabled this can expose the dangling read.

  Expected behavior: `from(std::string)` should keep the JSON input alive for as long as the returned `auto_parser` needs it.

  **simdjson release**

  Observed on the current main branch.

  **Configuration**

  Tested with C++20 and sanitizers enabled.

  Yes. I can provide a focused fix with regression coverage.

yangzhg:
Short title (summary):
Fix input lifetime in from(std::string)

Description
  `simdjson::from(std::string)` could leave the returned `auto_parser` referring to storage owned by the input string. Since on-demand conversion may happen after `from(...)` returns, temporary string input could be destroyed before the
  document is materialized.

  This change makes `auto_parser` own padded input for `std::string` inputs, so the backing storage remains valid for the parser/document lifetime. It also adds `padded_string&&` overloads so callers with already padded input can transfer
  ownership without copying the JSON payload.
  Issue reproduced / related issue: fixes #2703

Type of change
- [x] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

How to verify / test

Added regression tests covering:

  - `from(std::string)`
  - `from(parser, std::string)`
  - `from(std::move(padded_string))`
  - `from(parser, std::move(padded_string))`
 
  Verified locally with:

  ```bash
  cmake --build /tmp/simdjson-review-build-cxx20-asan --target ondemand_convert_tests -j 6
  ctest --test-dir /tmp/simdjson-review-build-cxx20-asan --output-on-failure -R ondemand_convert_tests
  cmake --build /tmp/simdjson-review-build-dev --target ondemand_convert_tests -j 6
  ctest --test-dir /tmp/simdjson-review-build-dev --output-on-failure -R ondemand_convert_tests


Checklist before submitting
- [x] Documentation / README updated if needed
- [x] I linked the related issue (if applicable)

Final notes
- For large PRs, prefer smaller incremental PRs or request staged review.

Thanks for the contribution!

### DIFF
--- a/include/simdjson/convert-inl.h
+++ b/include/simdjson/convert-inl.h
@@ -19,6 +19,12 @@ inline auto_parser<parser_type>::auto_parser(parser_type &&parser, padded_string
 }
 
 template <typename parser_type>
+inline auto_parser<parser_type>::auto_parser(parser_type &&parser, padded_string &&str) noexcept requires(!std::is_pointer_v<parser_type>)
+  : m_parser{std::move(parser)}, m_owned_input{std::move(str)}, m_doc{}, m_error{SUCCESS} {
+  m_error = m_parser.iterate(m_owned_input).get(m_doc);
+}
+
+template <typename parser_type>
 inline auto_parser<parser_type>::auto_parser(std::remove_pointer_t<parser_type> &parser, ondemand::document &&doc) noexcept requires(std::is_pointer_v<parser_type>)
   : m_parser{&parser}, m_doc{std::move(doc)} {}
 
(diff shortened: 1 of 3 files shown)


lemire:
Thanks. Your fix appears correct but it complexifies our API, and introduces a copy which we would rather not have.

The core issue is that our `std::string` overloads are garbage.

I am retiring them: https://github.com/simdjson/simdjson/pull/2705

(Note that I think that wrote the garbage code.)

(Note that `simdjson::from` is marked as experimental and subject to changes.)

PR #2698, opened 2026-04-25
builder: coalesce per-field separator+key+colon into one constexpr write

FranciscoThiesen:
## Summary

When the reflection-driven JSON serializer writes a struct, every non-first field
emits **three separate `string_builder::append` calls** (`,`, `"key"`, `:`),
each of which performs its own bounds check + small write. This PR coalesces
them into a single compile-time string per field so each field does **one
bounds check + one `memcpy`**.

Three call sites are patched in `include/simdjson/generic/builder/json_builder.h`
— `atom()` (~line 91), `append()` (~line 226), and `extract_from()` (~line 301).
All three have the same per-field pattern; `atom` is hit by every nested
struct, `append` by the top-level call from `to_json()` / `to_json_string()` /
`operator<<`, and `extract_from` by selective-field serialization.

Type of change
- [x] Optimization

## What the code did before (what each non-first field cost)

`include/simdjson/generic/builder/json_builder.h` (master, line 91):

```cpp
template for (constexpr auto dm : ...) {
  if (i != 0)
    b.append(',');                                       // capacity_check(1) + 1-byte write
  constexpr auto key = std::define_static_string(
    constevalutil::consteval_to_quoted_escaped(...));    // "\"foo\""
  b.append_raw(key);                                     // capacity_check(N) + memcpy(N)
  b.append(':');                                         // capacity_check(1) + 1-byte write
  atom(b, t.[:dm:]);
  i++;
}
```

`string_builder::append(char)` and `string_builder::append_raw(string_view)` each
call `capacity_check` (a `position + n <= capacity` test, possibly followed by
a `grow_buffer` call) and then write — see `json_string_builder-inl.h` lines
491-540 and 802-815. So three field-write calls = **three bounds checks** plus
one branch on `i != 0` per non-first field.

## What the code does after

```cpp
template for (constexpr auto dm : ...) {
  constexpr auto first_key = std::define_static_string(
      constevalutil::consteval_to_quoted_escaped(std::meta::identifier_of(dm)) + ":");
  constexpr auto rest_key = std::define_static_string(
      std::string(",") + constevalutil::consteval_to_quoted_escaped(std::meta::identifier_of(dm)) + ":");
  b.append_raw(i == 0 ? first_key : rest_key);
  atom(b, t.[:dm:]);
  i++;
}
```

The `","` + key + `":"` are concatenated at `consteval` time into a single
`define_static_string` per field, so the runtime sees **one `append_raw` per
[...]

PR #2706, opened 2026-05-02
builder: hoist write position into a local register in the struct atom

FranciscoThiesen:
## Summary

Reduce per-byte memory traffic in the reflection-driven struct serializer
by holding the write position in a local register across runs of fixed
bytes, instead of going through the `string_builder.position` member on
every write.

Layered on top of #2698 (already merged into master).

Type of change
- [x] Optimization

## The problem

After the per-field key coalescing in #2698, every byte the struct atom
wrote still went through `string_builder::position`. Because the buffer
itself is a `char*`, C++'s strict aliasing rules force the compiler to
assume that any write to `buffer.get()[...]` could overwrite `position`
or `capacity` — those are other members of the same object reachable
through the `string_builder &`. Result: after each byte/memcpy, the
compiler had to reload both fields from memory before the next write.

A representative inner-loop snippet from `atom<CITMSeatCategory>` writing
a `,` between vector elements (baseline post-#2698):

```
ldr   x0, [x24]              ; reload buffer.get()
add   x1, x22, #0x1          ; new position
str   x1, [x24, #8]          ; store position back
mov   w1, #0x2c              ; ','
strb  w1, [x0, x22]          ; write the comma
ldp   x22, x27, [x24, #8]    ; reload position AND capacity
cmp   x27, x22               ; check capacity for next byte
b.eq  <fallback>
ldr   x0, [x24]              ; reload buffer.get() AGAIN
...
```

Six instructions per single-byte write, of which only the `strb` is the
real work.

## What changed

Three small "unsafe" helpers added to `string_builder`:

- `bool reserve(size_t n)`: ensures at least `n` bytes are available,
  growing the buffer if needed.
- `char *unsafe_data()` and `size_t &unsafe_position()`: expose the
  buffer pointer and position so the caller can write directly.
- `void unsafe_append_uint(UInt v)`: writes the digits of an unsigned
  integer at the current position with no bounds check (caller has
  already reserved `MAX_UINT64_DIGITS`).

The reflection struct atom is rewritten to:

1. `reserve(N)` enough capacity for the next run of fixed bytes.
2. Pull the buffer pointer (`buf`) and write position (`pos`) into
   local variables.
3. Issue every write in that run directly through `buf[pos++]` or
   `memcpy(buf + pos, ..., N); pos += N`. No bounds checks, no member
   access — the compiler can prove the locals don't alias each other and
   keeps `pos` in a register.
4. Sync `pos` back to `b.unsafe_position()` and reload after any
   call that may grow the buffer or move position (`reserve` calls
   that may realloc, `unsafe_append_uint`, recursive `atom()` for
   variable-size fields).

Why the sync/reload dance is needed: `reserve` may `grow_buffer`
(realloc), which moves `buf`. Recursive `atom()` for strings,
vectors, and nested structs writes through `b.position` itself, so
our local `pos` and `b.position` must agree at the boundary.

The win is concentrated in the struct atom because that's where you have
runs of small fixed-byte writes (separator + key + colon + arithmetic
value) between calls. Container atoms have no such runs — just a comma
between recursive calls — and would not benefit, so they are unchanged.

## Performance

Build: gcc-trunk g++ 16.0.1, `-O3 -DNDEBUG -freflection -std=c++26`,
on aarch64 Linux. Median of 5 contemporaneous runs, both states built
clean on top of current master, output byte-identical to baseline.

CITM serialization (496 682 bytes):

| variant | baseline | patched | delta |
|---|---|---|---|
| simdjson_static_reflection | 3243 MB/s | 3831 MB/s | **+18.1%** |
| simdjson_reuse_buffer | 3422 MB/s | 3970 MB/s | **+16.0%** |
| simdjson_to | 3048 MB/s | 3593 MB/s | **+17.9%** |
| simdjson_to_reuse | 2992 MB/s | 3631 MB/s | **+21.4%** |

Twitter serialization (81 927 bytes):

| variant | baseline | patched | delta |
|---|---|---|---|
| simdjson_static_reflection | 5485 MB/s | 5660 MB/s | +3.2% |
| simdjson_reuse_buffer | 6774 MB/s | 6977 MB/s | +3.0% |
| simdjson_to | 5037 MB/s | 5066 MB/s | +0.6% |
| simdjson_to_reuse | 4986 MB/s | 5067 MB/s | +1.6% |

CITM benefits more than Twitter because its structs are smaller-fielded
(CITMArea has 2, CITMPrice has 3, CITMSeatCategory has 2) and those
fields are mostly arithmetic, so the per-field overhead dominates.
Twitter's structs have larger string fields, where the variable-size
write paths still dominate the cost.

## How to verify / test

```bash
cmake -B buildreflect -D SIMDJSON_STATIC_REFLECTION=ON -D SIMDJSON_DEVELOPER_MODE=ON
cmake --build buildreflect --target benchmark_serialization_citm_catalog \
    benchmark_serialization_twitter static_reflection_comprehensive_tests
```

The reflection round-trip tests in `static_reflection_comprehensive_tests`
exercise both the new struct atom and the unchanged `extract_from` and
pass with this change.

## AI usage disclosure

### DIFF
--- a/include/simdjson/generic/builder/json_builder.h
+++ b/include/simdjson/generic/builder/json_builder.h
@@ -89,14 +89,18 @@ template <class T>
            !std::is_same_v<T, const char*> &&
            !std::is_same_v<T, char> && !require_custom_serialization<T>)
 constexpr void atom(string_builder &b, const T &t) {
+  // Coalesce the per-field separator+key+colon writes into a single
+  // append_raw, so each field does one capacity_check + one memcpy instead of
+  // three. The leading-comma variant is selected at runtime by the compile-time
+  // peeled `i` counter, which clang folds away after the template-for unroll.
   int i = 0;
   b.append('{');
   template for (constexpr auto dm : std::define_static_array(std::meta::nonstatic_data_members_of(^^T, std::meta::access_context::unchecked()))) {
-    if (i != 0)
-      b.append(',');
-    constexpr auto key = std::define_static_string(constevalutil::consteval_to_quoted_escaped(std::meta::identifier_of(dm)));
-    b.append_raw(key);
-    b.append(':');
+    constexpr auto first_key = std::define_static_string(
+        constevalutil::consteval_to_quoted_escaped(std::meta::identifier_of(dm)) + ":");
+    constexpr auto rest_key = std::define_static_string(
+        std::string(",") + constevalutil::consteval_to_quoted_escaped(std::meta::identifier_of(dm)) + ":");
+    b.append_raw(i == 0 ? first_key : rest_key);
     atom(b, t.[:dm:]);
     i++;
   };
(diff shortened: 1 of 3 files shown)


FranciscoThiesen:
Abandoning this PR. Re-measured A/B in the standard p2996 Docker env (clang-p2996 21.0.0git, Release, aarch64 Linux) and CITM regresses by ~5% with the position-hoisting patch — opposite of what the PR description claims. The PR's +18% CITM number was measured with gcc-trunk g++ 16.0.1, which isn't reproducible in the project's standard build environment. Will revisit with a different approach.

PR #2707, opened 2026-05-03
builder: force-inline atom templates and replace integer writer

FranciscoThiesen:
## Summary

Speed up reflection-driven JSON serialization by **~43% on integer-heavy
structs (CITM), ~19% on string-heavy structs (Twitter), and ~18% on the
signed-integer write path** via two complementary changes to the builder
headers, plus a dead-code cleanup. Output is byte-identical to master.

Type of change
- [x] Optimization
- [x] Cleanup (~80 lines of unused integer-formatting helpers removed)

## The problem

In `string_builder`, the underlying buffer is a `std::unique_ptr<char[]>`,
so every byte the serializer writes goes through a `char*`. C++'s strict
aliasing rules say a write through `char*` may alias *any* object —
including the `size_t position` and `size_t capacity` members of the same
`string_builder`. The compiler must therefore assume those members can be
clobbered by every byte write, and reload them from memory before the next
operation.

This penalty is unavoidable when neighboring atom calls aren't inlined,
because the compiler can't see across the call boundary. Looking at a
baseline build's symbol table, the integer atom remained a real
out-of-line function:

```
nm --print-size --demangle .../benchmark_serialization_citm_catalog
...
460 bytes  void simdjson::arm64::builder::atom<unsigned long, void>(...)
152 bytes  fast_digit_count_64::table
```

Every uint64 field went through this function call. The function-call
overhead plus the spill/reload of `position`/`capacity` around each call
dominated CITM's profile (~38% of samples in the integer atom alone).

## What changed

**(1) Force-inline the hot `atom<T>` overloads.** Add `simdjson_really_inline`
(= `inline __attribute__((always_inline))`) to:

- `atom<arithmetic>`
- `atom<struct>` (the `template for`-driven one)
- `atom<container_but_not_string>` (begin/end iterator variant)
- `atom<appendable_containers>`
- `atom<optional_type>`
- `atom<string-like>`

Previously these were `constexpr` (required for `template for`) but
inlining was just a hint — the compiler routinely chose to leave the
non-trivial instantiations as separate functions.

**(2) Replace the integer writer with a forward cascade-on-magnitude
implementation, applied to both the unsigned and signed branches.**

The old code computed `digit_count(v)` upfront, then wrote backward in
4-digit batches via a loop:

```cpp
size_t dc = internal::digit_count(pv);            // function call
char *write_pointer = buffer.get() + position + dc - 1;
while (pv >= 10000) { ... write backward ... }
// tail handlers for remaining 1-4 digits
position += dc;
```

The new code is a straight-line cascade — `if (v < 100) … else if (v < 10000)
… else if (v < 100000000) …` — that writes digits *forward* without sizing
the number first. No loop, no helper call. The decimal-pair lookup table
is unchanged. The writer is built from a non-recursive DAG of
`always_inline` helpers (`write_lt100`, `write_lt10000`, `write_4_digits`,
`write_lt1e8`, `write_uint_jeaiii`) — gcc and MSVC reject `always_inline`
on recursive functions, so the implementation strictly composes
smaller-domain helpers without self-calls.

The signed branch handles the sign branchlessly: it always writes `'-'`
at `position` and advances position only when `v < 0`, then calls
`write_uint_jeaiii` on the absolute value. With both branches now using
the same writer, `int_log2`, `fast_digit_count_32`, `fast_digit_count_64`,
and `digit_count` (~80 lines) are removed as dead code.

## Why both — and why they're non-linear

Each change in isolation produces only a small gain (force-inline alone is
within noise; the forward writer alone gives ~+10% on CITM). The
combination produces +43%. Three effects interact:

1. **Aliasing collapse.** With `simdjson_really_inline` on the atom chain,
   the compiler sees the full inlined body of an entire struct
   serialization as one straight-line sequence of writes against a single
   `string_builder&`. With no opaque calls in between, it can hoist
   `b.position` into a register, run the entire struct as a local, and
   only flush back to memory at the end.

2. **Capacity-check fusion.** Once everything is inlined, consecutive
   `capacity_check(N)` calls become visible to the optimizer. It can
   fold redundant checks and combine the remaining ones into a single
   check.

3. **The straight-line writer is what enables (1) and (2).** The OLD
   writer had a loop (`while (pv >= 10000)`) and a `digit_count` helper
   call. Loops resist cross-call inlining: even when the wrapping atom is
   forced inline, the loop body is typically left as a separate code
   region. The NEW cascade is straight-line `if/else if` with direct
   `std::memcpy` calls, which the compiler readily integrates into the
   surrounding struct atom's machine code.

So change (1) alone leaves the loop blocking the inliner, and change (2)
alone leaves the wrapping `atom<T>` opaque. Together: door open AND
nothing on the other side blocking — `atom<CITMPrice>` (3 uint64 fields)
compiles into ~30 lines of straight-line assembly with one register
holding `position` and a single batched capacity check.

The symbol-table difference is concrete:

```
BASELINE:  atom<unsigned long>           460 bytes (separate function)
           fast_digit_count_64::table    152 bytes
PATCHED:   write_uint_jeaiii             ~340 bytes (single helper)
           atom<unsigned long>           --- (fully inlined)
           fast_digit_count_64::table    --- (deleted)
```

## Performance

Build: clang-p2996 21.0.0git, `-O3 -DNDEBUG`, Release, on aarch64 Linux
(Docker on Apple Silicon). Both binaries built fresh in the same docker
invocation. Median of 7 alternating rounds.

CITM Catalog (496 682 bytes output):

| | min | median | max |
|---|---|---|---|
| baseline | 2971 MB/s | 3230 MB/s | 3399 MB/s |
| **patched** | **4599 MB/s** | **4615 MB/s** | **4755 MB/s** |

→ **+42.9%**

Twitter (81 927 bytes output):

| | min | median | max |
|---|---|---|---|
| baseline | 5340 MB/s | 5403 MB/s | 5438 MB/s |
| **patched** | **6341 MB/s** | **6453 MB/s** | **6481 MB/s** |

→ **+19.4%**

Signed-int microbench (200 × 100 000 mixed-sign mixed-magnitude `int64_t`
values, written through `string_builder::append(int64_t)` plus a comma):

| | min | median | max |
|---|---|---|---|
| baseline | 2251 MB/s | 2272 MB/s | 2298 MB/s |
| **patched** | **2672 MB/s** | **2685 MB/s** | **2769 MB/s** |

→ **+18.2%**

(Microbench is one-shot for verification only and is not part of the diff.)

CITM benefits more than Twitter because its records are mostly small
integer fields (CITMPrice: 3 × uint64; CITMArea: uint64 + vector<uint64>;
etc.). Twitter has more strings, where the existing SIMD escape path was
already efficient and the integer-write speedup is a smaller fraction of
total time. The signed-int microbench is dominated by the integer writer
itself (no struct atom layer), so the gain reflects the per-integer
write-path improvement plus the new branchless sign handling.

Output is **byte-identical** to baseline on both inputs (CITM 496 682
bytes, Twitter 81 927 bytes). The reflection round-trip tests in
`static_reflection_comprehensive_tests` (which exercise both signed and
unsigned integer fields) all pass.

## How to verify / test

```bash
cmake -B buildreflect -D SIMDJSON_STATIC_REFLECTION=ON -D SIMDJSON_DEVELOPER_MODE=ON
cmake --build buildreflect --target benchmark_serialization_citm_catalog \
    benchmark_serialization_twitter static_reflection_comprehensive_tests
```

## AI usage disclosure

### DIFF
--- a/include/simdjson/generic/builder/json_builder.h
+++ b/include/simdjson/generic/builder/json_builder.h
@@ -23,7 +23,7 @@ namespace builder {
 
 template <class T>
   requires(concepts::container_but_not_string<T> && ! concepts::optional_type<T> && !require_custom_serialization<T>)
-constexpr void atom(string_builder &b, const T &t) {
+simdjson_really_inline constexpr void atom(string_builder &b, const T &t) {
   auto it = t.begin();
   auto end = t.end();
   if (it == end) {
@@ -45,7 +45,7 @@ template <class T>
            std::is_same_v<T, std::string_view> ||
            std::is_same_v<T, const char *> ||
            std::is_same_v<T, char>)
-constexpr void atom(string_builder &b, const T &t) {
+simdjson_really_inline constexpr void atom(string_builder &b, const T &t) {
   b.escape_and_append_with_quotes(t);
 }
 
@@ -74,7 +74,7 @@ constexpr void atom(string_builder &b, const T &m) {
 
 template<typename number_type,
          typename = typename std::enable_if<std::is_arithmetic<number_type>::value && !std::is_same_v<number_type, char>>::type>
-constexpr void atom(string_builder &b, const number_type t) {
+simdjson_really_inline constexpr void atom(string_builder &b, const number_type t) {
   b.append(t);
 }
 
@@ -88,7 +88,7 @@ template <class T>
            !std::is_same_v<T, std::string_view> &&
            !std::is_same_v<T, const char*> &&
            !std::is_same_v<T, char> && !require_custom_serialization<T>)
-constexpr void atom(string_builder &b, const T &t) {
+simdjson_really_inline constexpr void atom(string_builder &b, const T &t) {
   // Coalesce the per-field separator+key+colon writes into a single
   // append_raw, so each field does one capacity_check + one memcpy instead of
   // three. The leading-comma variant is selected at runtime by the compile-time
@@ -110,7 +110,7 @@ constexpr void atom(string_builder &b, const T &t) {
 // Support for optional types (std::optional, etc.)
 template <concepts::optional_type T>
   requires(!require_custom_serialization<T>)
-constexpr void atom(string_builder &b, const T &opt) {
+simdjson_really_inline constexpr void atom(string_builder &b, const T &opt) {
   if (opt) {
     atom(b, opt.value());
   } else {
@@ -156,7 +156,7 @@ template <concepts::appendable_containers T>
            !concepts::optional_type<T> && !concepts::smart_pointer<T> &&
            !std::is_same_v<T, std::string> &&
            !std::is_same_v<T, std::string_view> && !std::is_same_v<T, const char*> && !require_custom_serialization<T>)
-constexpr void atom(string_builder &b, const T &container) {
+simdjson_really_inline constexpr void atom(string_builder &b, const T &container) {
   if (container.empty()) {
     b.append_raw("[]");
     return;
--- a/include/simdjson/generic/builder/json_string_builder-inl.h
+++ b/include/simdjson/generic/builder/json_string_builder-inl.h
@@ -638,55 +638,74 @@ static const char decimal_table[200] = {
 };
 
 // Forward unsigned-int writer (cascade-on-magnitude, no upfront digit_count).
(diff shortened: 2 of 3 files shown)


lemire:
Verified manually.

PR #2708, opened 2026-05-06
builder: position-as-local writer for reflection serializer (+55% CITM)

FranciscoThiesen:
## Summary

Refactor the reflection serialization atom path to keep the write
position in a CPU register across the entire inlined call chain,
instead of as a member of `string_builder`. Output is byte-identical
to master.

| benchmark | baseline | patched | gain |
|---|---|---|---|
| **CITM Catalog** (496 682 bytes) | 4696 MB/s | **6806 MB/s** | **+44.9%** |
| **Twitter** (81 927 bytes)        | 6456 MB/s | 6381 MB/s     | within noise |

Type of change
- [x] Optimization

Default initial capacity is unchanged at 1024. Public `string_builder`
API is unchanged.

## The problem

In master, the reflection atom code writes through
`string_builder::buffer.get()` and updates `string_builder::position`
after every write. Because `buffer.get()` returns a `char *` and the
strict aliasing rules let a `char *` write alias *any* object —
including the `size_t position` and `size_t capacity` members of the
*same* `string_builder` instance — the compiler is forced to reload
both members from memory after every byte the serializer emits.

`perf annotate` on a `RelWithDebInfo` build of master
(`benchmark_serialization_citm_catalog -f simdjson_static_reflection`)
shows this directly. The dominant function (`append<CitmCatalog>`)
contains many copies of this pattern, each accounting for 4-5% of
total CPU time:

```
        : 495   if (simdjson_likely(upcoming_bytes <= capacity - position)) {
   4.29 :   296bc:  ldr     x8, [x28]                  ; reload `capacity`
---
   0.00 :   296d8:  strb    w9, [x8, x21]              ; write a byte at buffer[position]
   4.53 :   296dc:  ldur    x8, [x29, #-8]             ; reload `buffer` pointer from stack
---
   0.01 :   29714:  ldr     x21, [x19, #8]             ; reload `position` (offset 8 in string_builder)
   5.09 :   29718:  ldr     x8, [x28]                  ; reload `capacity`
---
   0.25 :   29738:  ldr     x21, [x19, #8]             ; reload `position`
   4.47 :   2973c:  ldr     x8, [x28]                  ; reload `capacity`
---
   0.37 :   29778:  ldr     x8, [x28]                  ; reload `capacity`
   5.44 :   2977c:  ldr     x24, [x20]                 ; reload (vector iterator)
---
        : 495   if (simdjson_likely(upcoming_bytes <= capacity - position)) {
   5.48 :   29bd8:  ldr     x8, [x28]                  ; reload `capacity`
---
   0.09 :   29c1c:  ldr     x22, [x28]                 ; reload
   5.41 :   29c20:  ldr     x21, [x15]                 ; reload
```

Adding the visible reload pairs in the perf annotate output gives
~25-30% of `append<CitmCatalog>` CPU time spent reloading state from
memory — not the actual byte-level write work. These reloads are not
redundant from the compiler's point of view: each preceding `strb`
*could* legally have clobbered the size_t members (a write through
`char *` is allowed to alias anything). The compiler is correct to
reload; the design is what's forcing it to.

## The fix

Move the write position from being a member of `string_builder` to a
local `size_t` for the duration of an `append()` call.

A new internal `writer` struct (in `json_builder.h`, internal namespace)
holds:

- `char *ptr` — buffer pointer (refreshed only after a grow)
- `size_t pos` — write position (purely local)
- `size_t cap` — capacity (refreshed only after a grow)
- `string_builder *sb` — back-reference for grow / final sync

The reflection atom family is rewritten to take `writer &` instead of
`string_builder &`. The top-level entry point `append(string_builder &b,
const T &t)` constructs a `writer` on the stack from `b`, threads it
through the inlined atom chain, then syncs `pos` back into
`b.position` once at the end:

```cpp
template <class T>
simdjson_inline void append(string_builder &b, const T &t) {
  writer w(b);
  atom(w, t);   // fully inlined chain, all writes through w
  w.sync();
}
```

Because `writer` is a true stack-local in `append()` and every `atom`
overload is `simdjson_really_inline`, after SROA the three fields
become register-resident across the entire serialization. The writes
in atom now look like:

```cpp
if (!w.ensure(N)) return;
w.ptr[w.pos++] = c;        // pos in a register; no reload
```

The compiler can keep `w.pos` in a register across hundreds of writes
because there is *no memory location for it* — `w.pos` only hits memory
when `sync()` is called at the very end (or before the slow
`grow_slow` path on the rare reallocation).

The string-escape path inside the atom for `string` / `string_view` /
`char` / map-key fields is similarly inlined through the writer:
the existing `write_string_escaped(input, out)` helper already takes
a destination `char *` and returns bytes-written, so it composes
cleanly with `w.ptr + w.pos` without any sync/reload.

## Generated-code evidence

The change is visible in the perf profile shape. Top symbols measured
on the same input on the same machine in the same docker invocation:

```
master (baseline):
   77.97%  append<CitmCatalog>(string_builder&, ...)         (giant inlined fn)
   10.88%  atom<map<string, CITMEvent>>(string_builder&, ...) (un-inlined map iter)
    8.59%  libc.so.6 memcpy
    0.42%  libc.so.6 (memcpy variant)
    0.37%  libc.so.6 (memcpy variant)

patched (this PR):
   81.40%  bench_simdjson_static_reflection lambda           (everything inlined into the bench)
   12.16%  libc.so.6 memcpy
    2.71%  write_string_escaped                              (string-escape leaf, expected)
    0.64%  libc.so.6 (memcpy variant)
    0.50%  memcpy@plt
```

Two structural changes are visible:

1. The 10.88% un-inlined `atom<map>` symbol disappears entirely in
   the patched build (the map atom now inlines through the writer
   along with everything else).
2. The dominant function changes shape. The patched profile's
   `perf annotate` no longer shows the `ldr [x19, #8]` /
   `ldr [x28]` reload pattern after byte writes; the top hot lines
   move into actual work — `strh w9, [x8, #6]` (writing a digit pair
   from the jeaiii decimal table) at 6.26%, and vector-iterator setup
   at 3.25%. The *bookkeeping* cost has been replaced by *write* cost.

In other words, the bottleneck before this change was **the
serializer reloading its own state from memory after every byte
written**. The bottleneck after this change is the actual byte-level
work (integer formatting + string copies).

## Why CITM benefits much more than Twitter

The reload pattern's cost per record scales with the number of small
writes per struct, not with the byte count of the output. CITM's
records are dense in fixed-size integer fields — each contains many
short writes (separator / key / colon / one or two uint64 digits /
repeat) between which the reload pattern would fire. A typical CITM
`Performance` has 9 fields including 3 × `uint64_t` plus several
optional strings; a `Price` has 3 × `uint64_t`. Across ~1800 nested
struct invocations per CITM serialization, that is a lot of reload
pairs.

Twitter's structs are dominated by string fields with text bodies
that get copied as a single (sometimes long) memcpy. For string-heavy
records the reload pattern fires far less often per byte of output —
most of the per-byte cost lives inside the SIMD escape path, which
this PR doesn't change. The result is no measurable Twitter
regression and no significant Twitter win, while CITM picks up +45%.

This shape — the architectural fix is dramatic on integer-dense
structs, neutral on string-dense ones — is exactly what the diagnosis
predicted from the perf annotate output.

## Performance

Build: clang-p2996 21.0.0git, `-O3 -DNDEBUG`, on aarch64 Linux
(Docker on Apple Silicon). Both binaries built fresh from clean
build directories in the *same* docker invocation, RelWithDebInfo so
perf annotate can resolve symbols. Median of 7 alternating rounds.

CITM Catalog (496 682 bytes output):

| | min | median | max |
|---|---|---|---|
| baseline | 4614 MB/s | 4696 MB/s | 4849 MB/s |
| **patched** | **6713 MB/s** | **6806 MB/s** | **7041 MB/s** |

→ **+44.9%**. Patched run-to-run spread is 4.9% (6713–7041), which
fully clears the baseline range — every round of patched is faster
than every round of baseline.

Twitter (81 927 bytes output):

| | min | median | max |
|---|---|---|---|
| baseline | 6358 MB/s | 6456 MB/s | 6514 MB/s |
| **patched** | **6300 MB/s** | **6381 MB/s** | **6888 MB/s** |

→ within noise (-1.2% on the median; per-round patched ranges from
−4% to +6% of the baseline median, so this is not a measurable
change either way).

Output is **byte-identical** to baseline on both inputs (CITM 496 682
bytes, Twitter 81 927 bytes). All
`static_reflection_comprehensive_tests` round-trip tests pass —
they exercise integer / signed integer / float / boolean / string /
optional / vector / map / nested struct fields.

## Compatibility

- Public `string_builder` API is unchanged. Any user code that calls
  `sb.append(...)`, `sb.append_raw(...)`, etc. continues to work
  exactly as before.
- `DEFAULT_INITIAL_CAPACITY` is unchanged at 1024. This change is
  purely about the per-byte cost inside the inlined hot path.
- The `writer` struct is in the `simdjson::IMPLEMENTATION::builder`
  namespace and is intended for internal use only. The new
  `unsafe_data() / unsafe_position() / unsafe_capacity() /
  unsafe_set_position() / unsafe_grow()` accessors on `string_builder`
  are similarly internal hooks for the writer; they're public for
  visibility from `json_builder.h` (which is a separate header) but
  named `unsafe_*` to signal they should not be used directly.

## How to verify / test

```bash
cmake -B buildreflect -D SIMDJSON_STATIC_REFLECTION=ON \
                      -D SIMDJSON_DEVELOPER_MODE=ON
cmake --build buildreflect --target \
    benchmark_serialization_citm_catalog \
    benchmark_serialization_twitter \
    static_reflection_comprehensive_tests
```

## AI usage disclosure

### DIFF
--- a/include/simdjson/generic/builder/json_builder.h
+++ b/include/simdjson/generic/builder/json_builder.h
@@ -21,23 +21,107 @@ namespace simdjson {
 namespace SIMDJSON_IMPLEMENTATION {
 namespace builder {
 
+// Forward-declare helpers defined in json_string_builder-inl.h so the
+// writer-based atom code below can call them (the -inl.h is not yet
+// included at the point this header is parsed; without these forwards,
+// name lookup falls back to the wrong outer namespace).
+namespace internal {
+simdjson_really_inline char *write_uint_jeaiii(char *p, uint64_t v) noexcept;
+} // namespace internal
+inline size_t write_string_escaped(const std::string_view input, char *out);
+
+// =============================================================
+// `writer`: position-as-local hot-path writer used by the reflection
+// atom code below. Holds the buffer pointer, write position and
+// capacity in three fields that, once `writer` itself is a stack-local
+// in the caller and all atom() functions are inlined, become true
+// register-resident locals after SROA. Glaze achieves the same effect
+// by passing `B&& b, auto&& ix` through every helper. Holding `pos`
+// in a register (rather than as a member of string_builder) is what
+// breaks the strict-aliasing penalty on every char* write through the
+// buffer, which forces a reload of `b.position` and `b.capacity`
+// after every byte.
+// =============================================================
+struct writer {
+  char *ptr;        // buffer pointer (refreshed after a grow)
+  size_t pos;       // write position (local)
+  size_t cap;       // capacity (refreshed after a grow)
+  string_builder *sb;  // back-ref for grow / sync
+
+  // Snapshot string_builder state into a writer for the duration of
+  // a write chain.
+  simdjson_really_inline writer(string_builder &builder) noexcept
+      : ptr(builder.unsafe_data())
+      , pos(builder.unsafe_position())
+      , cap(builder.unsafe_capacity())
+      , sb(&builder) {}
+
+  // Write the local position back to the underlying string_builder.
+  // Caller is responsible for invoking before the writer is dropped
+  // (otherwise data is lost). Idempotent.
+  simdjson_really_inline void sync() noexcept {
+    sb->unsafe_set_position(pos);
+  }
+
+  // Ensure at least `n` more bytes of free capacity. Grows the
+  // underlying buffer if needed (rare path). Returns false on
+  // allocation failure.
+  simdjson_really_inline bool ensure(size_t n) noexcept {
+    if (simdjson_likely(pos + n <= cap)) return true;
+    return grow_slow(n);
+  }
+
+  // Slow path of ensure(). Out-of-line via simdjson_inline (not
+  // simdjson_really_inline) to keep the hot path short.
+  simdjson_inline bool grow_slow(size_t n) noexcept {
+    // Detect overflow.
+    if (simdjson_unlikely(pos + n < pos)) return false;
+    sb->unsafe_set_position(pos);
+    if (!sb->unsafe_grow((std::max)(cap * 2, pos + n))) {
+      return false;
+    }
+    ptr = sb->unsafe_data();
+    cap = sb->unsafe_capacity();
+    return true;
+  }
+};
+
+// === Helper: invoke a string_builder member that writes variable-length
+// content (escape_and_append_with_quotes etc), syncing the writer's local
+// state before the call and reloading after. Used for string fields where
+// rewriting the entire SIMD escape path through the writer would be a much
+// bigger refactor.
+template <class F>
+simdjson_really_inline void call_through_string_builder(writer &w, F &&f) noexcept {
+  w.sync();
+  f(*w.sb);
+  w.ptr = w.sb->unsafe_data();
+  w.pos = w.sb->unsafe_position();
+  w.cap = w.sb->unsafe_capacity();
+}
+
 template <class T>
   requires(concepts::container_but_not_string<T> && ! concepts::optional_type<T> && !require_custom_serialization<T>)
-simdjson_really_inline constexpr void atom(string_builder &b, const T &t) {
+simdjson_really_inline constexpr void atom(writer &w, const T &t) {
   auto it = t.begin();
   auto end = t.end();
   if (it == end) {
-    b.append_raw("[]");
+    if (!w.ensure(2)) return;
+    std::memcpy(w.ptr + w.pos, "[]", 2);
+    w.pos += 2;
     return;
   }
-  b.append('[');
-  atom(b, *it);
+  if (!w.ensure(1)) return;
+  w.ptr[w.pos++] = '[';
+  atom(w, *it);
   ++it;
   for (; it != end; ++it) {
-    b.append(',');
-    atom(b, *it);
+    if (!w.ensure(1)) return;
+    w.ptr[w.pos++] = ',';
+    atom(w, *it);
   }
-  b.append(']');
+  if (!w.ensure(1)) return;
+  w.ptr[w.pos++] = ']';
 }
 
 template <class T>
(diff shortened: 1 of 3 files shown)


lemire:
I love it. There are tweaks that I'd like to make, but I can do it as a follow up.

ZX41R:
I think this needs one safety fix before merge.

This PR drops the overflow guards that are currently in `string_builder::escape_and_append()` and `escape_and_append_with_quotes()`.

On `master`, a forged oversized `std::string_view` is rejected as `OUT_OF_CAPACITY`. On this branch, `2 + 6 * input.size()` wraps to a small value, `capacity_check()` succeeds, and `write_string_escaped()` walks past the actual backing buffer.

Minimal local reproducer:

```cpp
#include "simdjson.h"

#include <iostream>
#include <limits>
#include <string_view>

int main() {
  using simdjson::fallback::builder::string_builder;

  const char one_byte[] = "x";
  const size_t huge = ((std::numeric_limits<size_t>::max)() - 2) / 6 + 1;

  string_builder b(32);
  std::string_view forged(one_byte, huge);

  b.escape_and_append_with_quotes(forged);

  std::string_view out;
  auto err = b.view().get(out);
  std::cout << "err=" << err << " size=" << out.size() << "\n";
}
```

Built with:

```sh
c++ -std=c++20 -O0 -g -fsanitize=address,undefined -fno-sanitize-recover=all \
  -Iinclude -Isrc repro.cpp src/simdjson.cpp
```

`master`:

```text
err=OUT_OF_CAPACITY: The capacity was exceeded, we cannot allocate enough memory. size=0
```

This branch:

```text
ERROR: AddressSanitizer: stack-buffer-overflow
READ of size 16
  #0 simdjson::fallback::builder::find_next_json_quotable_character(...)
  #1 simdjson::fallback::builder::write_string_escaped(...)
  #2 simdjson::fallback::builder::string_builder::escape_and_append_with_quotes(...)
```

I would restore the two `std::numeric_limits<size_t>` checks in `json_string_builder-inl.h`.

There is a related wrap pattern in the new writer too:

```cpp
if (simdjson_likely(pos + n <= cap)) return true;
```

That should use the same shape as the existing hot path:

```cpp
if (simdjson_likely(n <= cap - pos)) return true;
```

and the slow path should compute `pos + n` only after checking `n <= max - pos`. That keeps the register-local writer optimization without reintroducing unsigned-capacity wraparound.

lemire:
@ZX41R Thanks.

PR #2716, opened 2026-05-08
Require physical padding for padded string views

ZX41R:
`padded_string_view::has_sufficient_padding()` currently treats trailing JSON whitespace inside the view as if it were physical padding after the view. That can make checked overloads such as `ondemand::parser::iterate(buf, len, len)` accept an input with `capacity == length` when the last 64 logical bytes are whitespace.

That is not safe for the On Demand iterator. If the buffer ends at a page boundary, the parser can return `SUCCESS`, then a later access can read into the next unmapped page.

This change makes `has_sufficient_padding()` require actual capacity beyond `length()`. It also updates `pad(std::string&)` so the returned view always excludes the full `SIMDJSON_PADDING` bytes from the logical length, including whitespace that was already present in the input string.

Added a regression test that maps two pages, protects the second page, places `{}` plus 64 trailing spaces at the end of the first page, and verifies that `iterate(buf, len, len)` returns `INSUFFICIENT_PADDING` instead of accepting the view.

Local checks:

```text
ninja -C build_asan ondemand_parse_api_tests
build_asan/tests/ondemand/ondemand_parse_api_tests

ninja -C build_asan ondemand_document_stream_tests ondemand_iterate_many_csv
build_asan/tests/ondemand/ondemand_document_stream_tests
build_asan/tests/ondemand/ondemand_iterate_many_csv

ninja -C build_asan padded_string_tests
build_asan/tests/padded_string_tests
```

All were run on an ASan/UBSan build.

### DIFF
--- a/include/simdjson/padded_string_view-inl.h
+++ b/include/simdjson/padded_string_view-inl.h
@@ -45,19 +45,7 @@ inline padded_string_view::padded_string_view(std::string_view s, size_t capacit
 }
 
 inline bool padded_string_view::has_sufficient_padding() const noexcept {
-  if (padding() >= SIMDJSON_PADDING) {
-    return true;
-  }
-  size_t missing_padding = SIMDJSON_PADDING - padding();
-  if(length() < missing_padding) { return false; }
-
-  for (size_t i = length() - missing_padding; i < length(); i++) {
-    char c = data()[i];
-    if (c != ' ' && c != '\t' && c != '\n' && c != '\r') {
-      return false;
-    }
-  }
-  return true;
+  return padding() >= SIMDJSON_PADDING;
 }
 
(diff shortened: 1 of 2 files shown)


lemire:
I am going to close this PR. 

The description is...

"That is not safe for the On Demand iterator. If the buffer ends at a page boundary, the parser can return SUCCESS, then a later access can read into the next unmapped page."

There is no explanation that follows. No example and as far as I can tell, no corresponding test. You have not said what is doing the access.


If you believe that there is an actual issue, please provide a clear explanation and, ideally,. a reproducible test case.

Your tests, as far as I can see, how tests your mental model. This is circular.

Your test needs to *demonstrate* the problem.

PR #2722, opened 2026-05-17
docs: exception-free API guide and HACKING.md in README

wyf027:
## Summary

- **#709**: Expands `doc/basics.md` under *Disabling exceptions* with On-Demand and DOM examples using `simdjson_result<T>` and explicit `.get()` / error checks when `SIMDJSON_EXCEPTIONS` is off.
- **#801**: Adds [HACKING.md](HACKING.md) to the README documentation list for contributors.

## Test plan

- [x] Documentation-only; examples match existing error-handling sections and Godbolt links.

### DIFF
--- a/doc/basics.md
+++ b/doc/basics.md
@@ -2253,6 +2253,49 @@ The simdjson can be build with exceptions entirely disabled. It checks the `__cp
 target_compile_definitions(simdjson PUBLIC SIMDJSON_EXCEPTIONS=OFF)
 ```
 
+#### API when exceptions are disabled
+
+When `SIMDJSON_EXCEPTIONS` is off, simdjson APIs return a `simdjson_result<T>`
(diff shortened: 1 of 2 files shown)


doc/basics.md
```diff
@@ -2253,6 +2253,49 @@ The simdjson can be build with exceptions entirely disabled. It checks the `__cp
 target_compile_definitions(simdjson PUBLIC SIMDJSON_EXCEPTIONS=OFF)
 ```
 
+#### API when exceptions are disabled
+
+When `SIMDJSON_EXCEPTIONS` is off, simdjson APIs return a `simdjson_result<T>`
```
lemire, line 2258: That's not quite correct. When `SIMDJSON_EXCEPTIONS` is set to zero, we only disable some throwing functions, but the API remains otherwise the same.

It is not a different API. We don't change what functions return.

PR #2727, opened 2026-05-17
perf(builder): add LSX fast_needs_escaping for LoongArch

### LINKED ISSUE #2603: Implement fast_needs_escaping for loongarch
Our string builder class has a simple fast_needs_escaping function for 64-bit ARM and x64 processors.

https://github.com/simdjson/simdjson/blob/e362b18499ef29543b828610777de8333f4ca53e/include/simdjson/generic/builder/json_string_builder-inl.h#L84C22-L84C41

We should add support for loongarch processors.

wyf027:
## Summary
- Add an LSX SIMD implementation of `fast_needs_escaping` for LoongArch (`__loongarch_sx`)
- Matches the existing LSX path for `find_next_json_quotable_character` in the same file

## Test plan
- [ ] CI on LoongArch (if available) or review by maintainers familiar with LSX intrinsics
- [x] Scalar fallback unchanged when LSX is not enabled

Fixes #2603

### DIFF
--- a/include/simdjson/generic/builder/json_string_builder-inl.h
+++ b/include/simdjson/generic/builder/json_string_builder-inl.h
@@ -173,6 +173,35 @@ simdjson_inline bool fast_needs_escaping(std::string_view view) {
   }
   return _mm_movemask_epi8(running) != 0;
 }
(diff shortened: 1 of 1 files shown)


lemire:
Running tests.

PR #2732, opened 2026-05-26
bound zero-skip loop to end in parse_decimal

jmestwa-coder:
The bounded parse_decimal skips fraction-leading zeros without a `p != end` guard, so a value like `0.000` whose zeros reach end reads one byte past it.


lemire:
Please demonstrate an issue with a reproducible test case.

I am going to close this.  Reopen if you believe that there is an issue, with an actual complete analysis and, ideally with a test.

PR #2742, opened 2026-05-30
Key selectors: compile-time perfect-hash field selection for On Demand

lemire:
**Currently experimental, will not merge as-is.**

## Summary

This PR adds **key selectors**, an experimental On-Demand feature for extracting a
known, fixed set of fields from a JSON object in a single pass, independently of
the order in which those fields appear in the document. This is joint work with @FranciscoThiesen 

You declare a selector type from a list of string literals and then walk an object
with a new `object::for_each<Selector>(callback)` method. Each selected key is
mapped at compile time to a small integer index; the callback is invoked once per
matching field with `(index, value)`, so the caller dispatches with a `switch`
instead of doing repeated string-keyed lookups.

```cpp
using namespace simdjson;
ondemand::parser parser;

using fields = ondemand::key_selector<"name", "city">;

std::string_view name, city;
  switch (i) {
    case 0: name = std::string_view(v); break; // "name"
    case 1: city = std::string_view(v); break; // "city"
  }
});
// name == "Daniel", city == "Montreal"  ("age" is skipped)
```

## Motivation

On Demand is fastest when fields are read in document order via `find_field`,
because each lookup resumes where the previous one stopped. When you do not know
(or do not control) the field order, the alternatives are less attractive:

- `find_field_unordered` / `operator[]` can rewind and rescan, and you still pay a
  string comparison per probed key;
- reading fields in the "wrong" order degrades to repeated rescans.

Key selectors give an order-independent, single-pass option: the object is walked
once, and each field's key is classified by a **compile-time perfect hash
function (PHF)** into either "one of my N keys" (yielding its index) or "not
mine" (skipped without being parsed). Recognizing a key reduces to: scan its
length, compute a hash from a couple of bytes, and do one length-and-bytes
comparison — branch-light and SIMD-accelerated, with all hash tables baked in as
`static constexpr` data the compiler can fully inline.

## Status and prior work

THIS WILL BE MARKED AS EXPERIMENTAL since I am unsure about the actual interface.  In particular, @the-moisrex came up with a few designs, see https://github.com/simdjson/simdjson/pull/2247 and I tried something at https://github.com/simdjson/simdjson/pull/2274 Unfortunately, these prior designs had a substantial performance penalty. I am hoping that the current design is better for performance while being still simple to use.

I am absolutely open to criticism.

### Public API (`include/simdjson/generic/ondemand/`)

- **`key_selector.h`** (new): `key_selector<Keys...>`, a stateless type templated
  on string-literal keys. It builds a gperf-style perfect hash at compile time
  (`consteval`) and exposes:
  - `static constexpr std::size_t size()` — number of keys;
  - `static constexpr std::string_view key_at(i)` — key text at index `i`;
  - `static std::size_t match_raw(raw_json_string)` — hash a raw JSON key,
    returning its index in `[0, N)` or `N` on a miss (SIMD-accelerated on
    NEON/SSE2, with a scalar fallback).
- **`object.h` / `object-inl.h`**: `object::for_each<Selector>(callback)`. Walks
  the object once, invoking `callback(index, value)` for each field whose key is
  in the selector, in JSON order. Duplicate keys are ignored after the first
  match; iteration stops once all `Selector::size()` keys have matched or the
  object ends. Returns a `simdjson::error_code` (`SUCCESS`, or the first error
  encountered). The value is consumed inside the callback, so there is no
  look-ahead iterator state.
- **`concepts.h`**: a `key_selector_type` concept.
- **`amalgamated.h` / `dependencies.h`**: wiring so `key_selector.h` is included
  in the right place and its editor-only includes are declared for the
  single-header amalgamation.

### Requirements

Key selectors use C++20 features (concepts and class-type non-type template
parameters) and are only compiled when `SIMDJSON_SUPPORTS_CONCEPTS` is defined.
Exceptions are **not** required: `for_each` reports errors via its returned
`error_code`.

### Documentation (`doc/basics.md`)

A new **"Key selectors"** section, marked experimental, covering the C++20
requirement, a short primer on compile-time perfect hashing, the API and its
semantics, and three worked examples:

1. selecting a few top-level fields;
2. selecting fields from a nested object obtained via `find_field`;
3. an advanced example where a selected value is itself an object, handled by a
   nested `for_each`.

The two existing "Performance tip" notes about reading attributes in document
order, and the "seek the keys in order" performance bullet, now cross-reference
the new section.

### Tests (`tests/ondemand/`)

`ondemand_object_find_field_tests.cpp` gains a `key_selector`/`for_each` subtest
plus three tests that mirror the documentation examples verbatim
(`key_selector_example_toplevel`, `key_selector_example_nested`,
`key_selector_example_inner_object`), so the documented code is compiled and its
results asserted. The test is registered in `tests/ondemand/CMakeLists.txt`.

### Benchmark (`benchmark/partial_tweets/`)

A new `simdjson_ondemand_key_selector` variant of the `partial_tweets` benchmark
demonstrates the feature on a realistic workload (extracting seven tweet fields
and two nested user fields), and is wired into `bench_ondemand.cpp`.

The new approach is fast on this benchmark, but not necessarily quite as fast the conventional approach. Yet the conventional On Demand approach expects you to query the keys in order. So the new approach is more robust, performance wise.
## Performance

On the `partial_tweets` benchmark (Apple arm64, median of repeated runs):

| Variant | Time | Notes |
| --- | --- | --- |
| hand-tuned ordered `find_field` | ~92 µs | fastest; relies on document order |
| `for_each` key selector | ~99–105 µs | order-independent, single pass |
| `simdjson_dom` | ~133 µs | reference |

The selector is within roughly 8–14% of hand-tuned ordered access while removing
the ordering assumption. An internal-iteration (`for_each`) design was chosen over
an iterator/range design because it avoids per-element look-ahead bookkeeping and
keeps value lifetime correct by construction; in A/B testing it was measurably
faster than the iterator form.

## Compatibility

- Additive only: no existing public API changes; everything new is guarded by
  `SIMDJSON_SUPPORTS_CONCEPTS` and compiles away in pre-C++20 builds.
- Experimental: the key-selector API may change in a future release.

## Verification

- Full `cmake --build` is green, including the single-header amalgamation step.
- `ondemand_object_find_field_tests` passes, including the three
  documentation-mirroring examples and the nested-object case.
- The `partial_tweets` key-selector benchmark passes its correctness check
  against the `simdjson_dom` reference.


FranciscoThiesen:
A few things I noticed while reviewing — two correctness items and one bit of dead code.

### 1. 32-char keys are matched by the scalar fallback but silently missed under NEON/SSE2

`scan_key_length` only reads two 16-byte blocks (offsets `0..31`), so a key whose closing quote sits at offset 32 is never found:

```cpp
if (m0 != 0) return ctz(m0);        // offset 0..15
if (m1 != 0) return 16 + ctz(m1);   // offset 16..31
return MaxKeyLen + 1;               // a 32-char key's quote (offset 32) lands here
```

`match_raw` then rejects it via `len > max_key_len` and returns a miss. The scalar fallback, however, does find it:

```cpp
for (std::size_t i = 0; i <= MaxKeyLen; ++i)
    if (p[i] == '"') return i;      // returns 32 for a 32-char key
```

Since `MaxKeyLen <= 32` is allowed (and the docs say "at most 32 characters"), a `key_selector` with a 32-character key matches in scalar builds but silently misses on x86/ARM — a platform-dependent result. Fix is either to tighten the limit to 31 (`static_assert` + docs), or to extend the scan to cover offset `max_key_len`.

### 2. Escaped JSON keys are silently skipped

Matching runs on the raw bytes (`rjs.raw()`) and `scan_key_length` stops at the first `"`, so a document key written with escapes — e.g. `"id"` for `id` — won't equal the literal selector bytes and is dropped. This differs from `find_field_unordered`, which matches through escapes. It's mentioned in `key_selector.h`, but not in the user-facing `object::for_each` docs (`object.h` / `basics.md`); worth surfacing there since it's a real divergence from normal On Demand behavior. (No false positives — the full length + byte compare prevents that — it's miss-only.)

### 3. Dead code

`pick_table_size<N>()` is defined but never used; `choose_table_size` is the one actually doing the work. It can be removed.

lemire:
@FranciscoThiesen Thanks. The escaped JSON is deliberate. The other two are bugs.

lemire:
Closing in favour of https://github.com/simdjson/simdjson/pull/2774

PR #2751, opened 2026-05-31
docs: document validation of skipped JSON values (#1640)

### LINKED ISSUE #1640: Document precisely what we do and don't validate
A (partial?) list for documentation of what we do and don't validate for skipped values:

Strings:
- We ALWAYS validate the boundaries of strings (that strings are properly closed, taking into account `\"`). This means that even if a string has some invalid stuff inside, we validate enough that they cannot affect anything else.
- We ALWAYS validate string characters (newlines and control characters being disallowed).
- We only validate the content of escape characters if you unescape a string (like using `get_string()` or `unescaped_key()`). i.e. `\p` is not allowed, and `\u` must be followed by hex digits.

Numbers/Booleans/Null:
- For the purposes of structure, we ALWAYS treat any sequence of non-whitespace (except { } [ ] : and ,) is a single number/boolean/null. This means we'll treat that `-0ab-10 trrrue` as two values without a comma between them, whether you actually use the values or not.
- We only validate the content of numbers, booleans and null if you actually use them.

Arrays/Objects:
- We ALWAYS validate that objects/arrays are closed before continuing to iterate, even if the objects/arrays are skipped.
- We only check for missing or extra `,` or `:` when you iterate or index an array or object.
- We only check that keys are strings when you iterate or index an array or object.
- We only validate remaining `,`/`:` in an array or object if you fully iterate it.
- We only validate that the closing `]` or `}` matches the opening one if you fully iterate the object/array.

Document:
- We ALWAYS validate that the JSON is not empty.
- We NEVER check if there is extra JSON after your document (honestly, that seems like a bug--if you fully iterate the document we should check that there is nothing after).

wyf027:
## Summary
Adds **What gets validated when you skip values** to `doc/basics.md`, summarizing structural checks for skipped strings, numbers, arrays, and objects in the On-Demand API.

Fixes #1640.

## Test plan
- [x] Markdown-only change

### DIFF
--- a/doc/basics.md
+++ b/doc/basics.md
@@ -514,7 +514,7 @@ in skipped content are not reported until you access those values.
 - Missing or extra `,` or `:`, key type checks, and matching of opening/closing `]`/`}` are
   validated only when you iterate or index the array or object, or when you fully iterate it.
 
-See also [On-Demand design notes](ondemand_design.md#validate-what-you-use) and
+See also [On-Demand design notes](ondemand_design.md) and
 [#1640](https://github.com/simdjson/simdjson/issues/1640).
 
 * **Extracting Values:** You can cast a JSON element to a native type:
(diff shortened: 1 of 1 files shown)


lemire:

PR #2754, opened 2026-05-31
docs: link tape structure doc from README (Fixes #951)

### LINKED ISSUE #951: Exposure to Raw Tape and Field Offset on Tape (Developer API)
There are two developer APIs will strongly benefit the performance of FishStore use case of simdjson.

1. It would be very nice if the whole tape can be portable. In other words, once a record is parsed, it can be stored and exported in raw tape form so that later we can directly read fields out without parsing structural again. So, I would suggest an API like `get_tape() -> std::string/std::string_ref` to expose raw tape structure.

2. There is a need to rule out hash collision in FishStore efficiently. Now, we extract the field out and materialized the field in optional PSF value area. It would be very nice to have a direct link (offset with size) to the raw value on tape so as to avoid copy. 

Dong

wyf027:
## Summary
Add a README entry pointing to `doc/tape.md`, which documents the internal DOM tape layout for contributors and advanced users.

Fixes #951

## Test plan
- [x] Docs-only change; link target `doc/tape.md` exists on `master`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ Usage documentation is available:
   how you can work with it.
 * [API](https://simdjson.github.io/simdjson/) contains the automatically generated API documentation.
 * [Compile-Time Parsing](doc/compile_time.md) presents our compile-time parsing function (C++26 only).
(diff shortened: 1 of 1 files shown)


lemire:

PR #2761, opened 2026-05-31
docs: link implementation selection to performance guide

wyf027:
## Summary
Point readers at performance tuning from the CPU dispatch overview.

## Test plan
- [x] Docs-only change.


lemire:

PR #2770, opened 2026-05-31
docs: use relative link to builder.md from dom.md

wyf027:
## Summary
Replace absolute GitHub URL with relative builder.md link.

## Test plan
- [x] Docs-only.


lemire:

PR #2773, opened 2026-05-31
docs: fix dom.md on-demand and builder links

wyf027:
## Summary
Docs-only improvement.

## Test plan
- [x] Docs-only.


lemire:

PR #2774, opened 2026-06-06
Key selectors: compile-time perfect-hash field selection for On Demand  and static deserialization

lemire:
Upgrade of [#2742](https://github.com/simdjson/simdjson/pull/2742). 

This PR adds key selectors, an On-Demand facility for extracting a known,
fixed set of fields from a JSON object in a single pass, independently of the
order in which those fields appear in the document. This is joint work with
@FranciscoThiesen.

It wires key selectors into the C++26 static-reflection deserializer by default in many cases (although the option remains to use the old in-order approach). Contrary to PR 2742, it improves performance in our benchmarks on two systems. Furthermore, it makes the performance more robust (in case the order of the keys is not consistent).

## Summary

You declare a selector type from a list of string literals and then walk an
object with a new `object::for_each<Selector>(callback)` method. Each selected key
is mapped at compile time to a small integer index; the callback is invoked once
per matching field with `(index, value)`, so the caller dispatches with a `switch`
instead of doing repeated string-keyed lookups.

```cpp
using namespace simdjson;
ondemand::parser parser;

using fields = ondemand::key_selector<"name", "city">;

std::string_view name, city;
  switch (i) {
    case 0: name = std::string_view(v); break; // "name"
    case 1: city = std::string_view(v); break; // "city"
  }
});
// name == "Daniel", city == "Montreal"  ("age" is skipped)
```

## Motivation

On Demand is fastest when fields are read in document order via `find_field`,
because each lookup resumes where the previous one stopped. When you do not know
(or do not control) the field order, the alternatives are less attractive:

- `find_field_unordered` / `operator[]` can rewind and rescan, and you still pay a
  string comparison per probed key;
- reading fields in the "wrong" order degrades to repeated rescans.

Key selectors give an order-independent, single-pass option: the object is walked
once, and each field's key is classified by a compile-time perfect hash
function (PHF) into either "one of my N keys" (yielding its index) or "not mine"
(skipped without being parsed). Recognizing a key reduces to: scan its length,
compute a hash from a couple of bytes, and do one length-and-bytes comparison —
branch-light and SIMD-accelerated, with all hash tables baked in as
`static constexpr` data the compiler can fully inline.

  ## Performance

  Measured with GCC 16 via the `gcc16` docker station (`-O3 -std=c++26 -freflection`),
  best of 5 interleaved runs. **"main"** is commit `98fde69` (the point this PR forks
  from, so the only difference measured is this PR); **"this PR"** is the `key_selector`
  path. Two machines:

  - **Ice Lake** — AVX-512 x86-64 server, GCC 16.0.1
  - **ARM** — Apple Silicon arm64 / NEON, GCC 16.1.0

  ### 1. Static-reflection deserialization — `doc.get<T>()`

  Full-struct deserialization of a reflected type. On `main` this issues one `obj[key]`
  lookup per member; this PR walks the object once with a compile-time perfect-hash
  `key_selector`. **Faster on both workloads and both architectures:**

  | Workload | Machine | main `obj[key]` | this PR `key_selector` | Speedup |
  |---|---|---:|---:|---:|
  | CITM (1.7 MB)    | Ice Lake | 2.47 GB/s | 2.59 GB/s | **+4.7 %** |
  | CITM (1.7 MB)    | ARM      | 3.57 GB/s | 4.14 GB/s | **+16.2 %** |
  | Twitter (631 KB) | Ice Lake | 4.50 GB/s | 4.74 GB/s | **+5.3 %** |
  | Twitter (631 KB) | ARM      | 5.37 GB/s | 5.47 GB/s | **+2.0 %** |

  ### 2. `partial_tweets` micro-extraction — `bench_ondemand`

  Extracts a fixed set of fields per tweet, where a hand-written, order-aware
  `find_field` chain (`simdjson_ondemand`) is the tough baseline. The general,
  order-independent `key_selector` matcher (new here) is a few percent behind because
  every visited field pays generic matching:

  | Machine | `find_field` (main, unchanged on PR) | `key_selector` (this PR) | Δ |
  |---|---:|---:|---:|
  | Ice Lake | 6.61 GB/s | 6.43 GB/s | **−2.7 %** |
  | ARM      | 6.78 GB/s | 6.50 GB/s | **−4.1 %** |

  The PR does not slow `find_field` (measured flat across branches); it *adds*
  `key_selector` as an order-independent option that costs ~3–4 % on this fixed-field,
  known-order workload.

  ### 3. `partial_tweets` under Apple Clang (arm64, native)

  For cross-compiler context, the same `partial_tweets` benchmark built with Apple
  Clang 17 natively on the same Apple Silicon machine (Release) runs faster in absolute
  terms than the GCC-16 docker toolchain, but shows the same small gap:

  | Variant | GB/s | Δ |
  |---|---:|---:|
  | `simdjson_ondemand` (find_field) | 7.57 GB/s | — |
  | `simdjson_ondemand_key_selector` | 7.31 GB/s | **−3.4 %** |

  ### Takeaway

  - **Full reflected deserialization (what the default governs): a consistent win on
    both Ice Lake and ARM (+2 % … +16 %) — so making `key_selector` the default is the
    right call.**
  - Fixed-field micro-extraction with a known order: a hand-tuned `find_field` chain
    still edges out the general matcher (−3 % … −4 %); the gap is small and
    order-independence is the feature's whole point, so the choice is left to the user.

## Status and prior work

@the-moisrex came up with a few designs, see #2247, and I tried
something at #2274. Those prior designs had a substantial performance penalty; the
current design aims to be better for performance while staying simple to use.
Criticism is very welcome.

### Public API (`include/simdjson/generic/ondemand/`)

- `key_selector.h` (new): `key_selector<Keys...>`, a stateless type templated
  on string-literal keys. It builds a gperf-style perfect hash at compile time
  (`consteval`, with a Hash-and-Displace fallback) and exposes:
  - `static constexpr std::size_t size()` — number of keys;
  - `static constexpr std::string_view key_at(i)` — key text at index `i`;
  - `static std::size_t match_raw(raw_json_string)` — hash a raw JSON key,
    returning its index in `[0, N)` or `N` on a miss (SIMD-accelerated on
    NEON/SSE2, with a scalar fallback), plus a `match_raw(p, len)` overload that
    consumes a precomputed key length (see optimizations below).
- `object.h` / `object-inl.h`: `object::for_each<Selector>(callback)`. Walks
  the object once, invoking `callback(index, value)` for each field whose key is
  in the selector, in JSON order. Duplicate keys are ignored after the first
  match; iteration stops once all `Selector::size()` keys have matched or the
  object ends. Returns a `simdjson::error_code` (`SUCCESS`, or the first error
  encountered). The value is consumed inside the callback, so there is no
  look-ahead iterator state.
- `concepts.h`: a `key_selector_type` concept.
- `amalgamated.h` / `dependencies.h`: wiring so `key_selector.h` is included
  in the right place and its editor-only includes are declared for the
  single-header amalgamation.

### Single-pass reflective deserialization (new in this revision)

The C++26 static-reflection deserializer (`doc.get<T>()` / `simdjson::from`) can
now route through a key selector built from the struct's members, walking each
object once with `object::for_each` instead of issuing one ordered/unordered
lookup per member. 

```cpp
// Compile this translation unit with -DSIMDJSON_USE_KEY_SELECTOR_REFLECTION=1
struct Tweet { uint64_t id; std::string text; uint64_t retweet_count; };

// Keys are NOT in declaration order, yet deserialization succeeds in one pass.
auto json = R"({ "retweet_count": 7, "id": 12345, "text": "hello" })"_padded;
Tweet t;
auto error = parser.iterate(json).get(t);
```

### `for_each` micro-optimizations

- Key length from the structural index (`value_iterator::field_key_with_length`):
  after `field_key()` the current structural token is the `:` that follows the
  key, so the closing quote is recovered by a single back-step from the colon
  (whitespace-tolerant) instead of a forward SIMD scan. The new
  `key_selector::match_raw(p, len)` overload consumes the precomputed length.
- `seen` bitset (`for_each`): the per-call "already matched" tracker is a
  compile-time bitset (one `uint64_t` for ≤64 keys) instead of a
  `std::array<bool, N>` — one-store init and register-resident bit ops instead of
  a byte array in memory.


FranciscoThiesen:
I reviewed this locally: read the full diff, then ran the new tests plus an adversarial suite (fuzzing `match_raw` against a naive reference matcher, randomized documents with duplicate/escaped keys, whitespace before `:`, the 31/32-char boundary, early-stop + parent resumption, unconsumed values) on NEON, SSE2 and the scalar fallback, plus the reflection path under GCC 16. The matching and iteration machinery held up everywhere — nice work. Two real issues found, posting them separately. Here is the first.

### `for_each` is `noexcept`, but the documented callback style throws → `std::terminate`

`object::for_each` is declared `noexcept`, yet the user callback runs *inside* that frame — and both the "Key selectors" examples in `doc/basics.md` and `benchmark/partial_tweets/simdjson_ondemand_key_selector.h` use throwing conversions (`std::string_view(v)`, `uint64_t(v)`) in the callback. On a type mismatch the exception hits the `noexcept` boundary and terminates the process. Repro is the doc's own first example with `"name": 123`:

```cpp
auto json = R"({ "name": 123, "city": "Montreal" })"_padded;
ondemand::parser parser;
using fields = ondemand::key_selector<"name", "city">;
std::string_view name, city;
try {
    switch (i) {
      case 0: name = std::string_view(v); break; // throws INCORRECT_TYPE
      case 1: city = std::string_view(v); break;
    }
  });
} catch (simdjson_error &) { /* never reached */ }
```

```
libc++abi: terminating due to uncaught exception of type simdjson::simdjson_error: INCORRECT_TYPE: The JSON element does not have the requested type.
```

This only fires on unexpected input, which is exactly when error handling matters. It is also a new hazard class relative to the existing `noexcept` API surface: `find_field` etc. return `simdjson_result`, so throwing conversions execute in the *user's* frame; here user code executes inside the library's `noexcept` frame.

Suggested fix: make it conditionally noexcept, e.g. `noexcept(std::is_nothrow_invocable_v<Func&, std::size_t, value>)` (or simply drop `noexcept`). Alternatively, require error_code-returning callbacks in the docs and update the examples — but conditional noexcept keeps the doc examples valid as written.

FranciscoThiesen:
### Second issue: the new default reflection path turns previously-valid structs into compile errors

Because the key-selector path is now the **default** for reflective deserialization, every reflected struct inherits `key_selector`'s compile-time limits: member (or renamed) keys at most 31 characters, at most 255 members, distinct keys, no quote/backslash in renamed keys.

Demonstrated under GCC 16 (`-std=c++26 -freflection`):

```cpp
struct Config {
  uint64_t number_of_simultaneous_connections = 0; // 35 chars
  uint64_t id = 0;
};
// ...
Config c;
auto err = doc.get(c);
```

On this PR that fails to compile:

```
include/simdjson/generic/ondemand/key_selector.h:1000:31: error: static assertion failed:
key_selector keys must be at most 31 characters long
```

while the same code compiles and works on `master` and with `-DSIMDJSON_DISABLE_KEY_SELECTOR_REFLECTION=1`. Long member names are not rare in generated or config-style structs, so this will break some users' builds on upgrade, and the assertion text doesn't tell them about the escape hatch.

Since eligibility is fully detectable at compile time, the nicest fix is an automatic per-type fallback: have the reflective `tag_invoke` check `if constexpr` whether all eligible member keys fit the selector limits (length ≤ 31, count ≤ 255, distinct, no forbidden characters) and silently use the existing ordered per-member path otherwise. Then the default never breaks anyone and the macro remains as a global opt-out.

At minimum: mention `SIMDJSON_DISABLE_KEY_SELECTOR_REFLECTION` in the static_assert message, and document the limits in the "Order-independent reflective deserialization" section of `doc/basics.md` (currently only the "Key selectors" section lists them).

FranciscoThiesen:
Smaller notes from the same review pass:

- **Stale PR description (two spots).** It mentions a `key_selector_type` concept in `concepts.h` that isn't in the diff — relatedly, `for_each`'s `Selector` template parameter is unconstrained, and constraining it with such a concept would give much better error messages for a wrong type. And the "Single-pass reflective deserialization" example says `-DSIMDJSON_USE_KEY_SELECTOR_REFLECTION=1`, while the implemented macro is the opt-out `SIMDJSON_DISABLE_KEY_SELECTOR_REFLECTION`.
- **Development-check parity.** `object::begin()` returns `OUT_OF_ORDER_ITERATION` under `SIMDJSON_DEVELOPMENT_CHECKS` when the iterator isn't at the object start; `for_each` has no such check and silently walks from the current position (e.g. after a prior `find_field` on the same object). Worth mirroring the guard.
- **Error-path asymmetry in `for_each`.** Internal errors (`field_key`, `field_value`, `skip_child`) call `it.abandon()`, but an error returned by the callback does not. If that's intentional (leave the iterator recoverable for the caller), a comment would help. Relatedly, a *void* callback that swallows a fatal sub-iteration error reaches `it.skip_child()` without the `is_open()` guard that `object_iterator::operator++` uses as a safety rail; hard to trigger in practice, cheap to mirror.
- **No `simdjson_result<object>::for_each` forwarding**, so error-code-style chains (`doc["x"].get_object()`) need an explicit `object` extraction before calling `for_each`, unlike most of the object API.
- Tiny: "They keys must be distinct" in the `object.h` doc comment; "This limitations could be eased" in `doc/basics.md`; in `compute_window` the inner `if (off > min_len) continue;` is redundant with the loop bound; and a stray blank line is added in `object-inl.h` near the `simdjson_result<object>::find_field` overloads.

FranciscoThiesen:
Note: Review was AI assisted by Claude (new Fable model)

lemire:
@FranciscoThiesen very useful.

lemire:
@FranciscoThiesen  Comments addressed.

lemire:
closing in favour of https://github.com/simdjson/simdjson/pull/2776

PR #2776, opened 2026-06-13
Compile-time key selectors: single-pass, order-independent field extraction

lemire:
## Summary

This adds **`ondemand::key_selector`** and a family of **`object::for_each`** overloads that
let you pull a fixed, known-at-compile-time set of fields out of a JSON object in a **single
pass**, regardless of the order the keys appear in the document.

A `key_selector<"k0", "k1", ...>` builds, at compile time, a perfect-hash classifier for its
keys. `object::for_each<Selector>(...)` then walks the object once; for every field it
classifies the key with the selector and dispatches to the handler for that key. Keys not in
the selector are skipped. The walk ends as soon as every selected key has matched.

There are three ways to spell the handlers, and they can be mixed one-per-key:

```cpp
using namespace simdjson;
ondemand::parser parser;

std::string_view name, city;
uint64_t age{};

// 1. Bind fields straight to variables (assigned via value::get). No lambdas.
auto error = obj.for_each<"name", "city", "age">(name, city, age);
if (error) { /* handle a malformed document or a type mismatch */ }

// name == "Daniel", city == "Montreal", age == 42
```

When a field needs custom logic (a nullable value, a nested object, a transform), drop in a
lambda for that key. Targets and lambdas mix freely, one per key:

```cpp
uint64_t id{};
twitter_user user{};

obj.for_each<"id", "user", "name">(
  id,                                                  // direct target
  [&](ondemand::value v) -> error_code {               // nested object
    ondemand::object u;
    SIMDJSON_TRY( v.get_object().get(u) );
    return u.for_each<"id", "screen_name">(user.id, user.screen_name);
  },
  name                                                 // direct target
);
```

A named selector type can be reused, and an index-based single-callback form is available for
shared-state or `switch`-style logic:

```cpp
using fields = ondemand::key_selector<"name", "city", "age">;

// per-key lambdas with a named selector
obj.for_each<fields>(
  [&](ondemand::value v){ name = std::string_view(v); },
  [&](ondemand::value v){ city = std::string_view(v); },
  [&](ondemand::value v){ age  = uint64_t(v); }
);

// single callback receiving the matched key's index in [0, size())
  switch (i) { case 0: /* name */ break; case 1: /* city */ break; case 2: /* age */ break; }
});
```

## Why

The ordered `find_field` chain is fast but couples your code to the document's key order, and
the order-independent `obj["k"]` lookups restart the scan per key. A key selector gives you
order independence **and** a single pass: each field's key is hashed once by a table that is
fully materialized at compile time (`static constexpr`), so the classification inlines to a few
instructions with no per-call setup.

The direct-binding form also makes the common "extract these fields into these variables" case
as short as it can be — no integer indices, no `switch`, no `using` alias — while the lambda and
index forms remain for anything that needs real logic.

## API

All of the following live in `include/simdjson/generic/ondemand/` and are guarded by
`SIMDJSON_SUPPORTS_CONCEPTS` (C++20).

### `key_selector<Keys...>`

- Compile-time perfect-hash classifier over up to **255** keys, each at most **63** bytes.
- `Selector::size()` is the key count; `Selector::match_raw(...)` returns the matched key's
  index in `[0, size())`, or `size()` on a miss.
- A single-byte "window" discriminator is detected at compile time and used when one exists,
  otherwise a length-aware perfect hash is used. Both are selected with `if constexpr`, so the
  unused path is compiled out.

### `object::for_each` (and the `simdjson_result<object>` forwarders)

Each argument is a **field handler**, exactly one per key:

- A **deserialization target** — any type `T` for which `value::get(T&)` works (the builtin
  scalar/string types, `object`/`array`/`value`, optionals, and your own `tag_invoke` types).
  The matched value is assigned into it. This is detected with the existing `deserializable`
  concept, so passing something that is neither a target nor a callable is a clean compile
  error.
- A **callable** taking `ondemand::value`. It may return `void` (run for side effects) or
  anything convertible to `error_code` (an `error_code`, or a `for_each_result` from a nested
  `for_each`), in which case its result is propagated.

Overloads:

- `for_each<Selector>(handler0, handler1, ..., handlerN)` — N must equal `Selector::size()`
  (compiler-enforced).
- `for_each<"k0", "k1", ...>(handler0, handler1, ...)` — same thing without a separate
  `key_selector` alias.
- `for_each<Selector>(callback)` where `callback` takes `(std::size_t index, value)` — the
  single-callback index form; also `void` or `error_code`-returning.

Overload selection between the index form and the per-key forms is driven by `requires`
clauses on invocability, so the compiler picks the right one and gives a clear error otherwise.
`noexcept` is computed per handler (nothrow-invocable for callables, nothrow-deserializable for
targets), so `for_each` is `noexcept` exactly when all of its handlers are.

### Result and stopping behavior

`for_each` returns a `for_each_result`:

```cpp
struct [[nodiscard]] for_each_result {
  error_code   error{SUCCESS};       // first error encountered, SUCCESS if none
  std::size_t  matched_count{0};     // number of distinct selected keys that matched
  constexpr operator error_code() const noexcept { return error; }
};
```

- It is **`[[nodiscard]]`**: `for_each` reports parse/type errors (from a matched value, a
  returning handler, or the structural walk) only through this result, so it must not be
  silently dropped. To deliberately ignore it, assign it to a variable or cast to `void`.
- The walk **stops at the first failure**: a handler that returns a non-`SUCCESS` `error_code`,
  a target whose `value::get` fails (e.g. a type mismatch), or — in exception-enabled builds —
  a handler that throws. The returned `error` is the first failure; an exception propagates to
  the caller.

### Usable without exceptions

The target form and the `error_code`-returning lambda form use only error codes, so the entire
API is usable in builds compiled with `SIMDJSON_EXCEPTIONS=0` / `-fno-exceptions`. In that
configuration the `for_each_result` is the sole channel for errors, which is why it is
`[[nodiscard]]`.

## Implementation notes

The per-key forms delegate to the single index-based `for_each<Selector>` via a small
`std::tuple` + `std::index_sequence` adapter (`object-inl.h`): the runtime matched-index is
turned into a compile-time selection among the handlers. The adapter runs once per *matched*
key (typically a handful), and the hot path — structural iteration, `match_raw`, the
single-byte window fast path, the "seen" bitset, and building a `value` only for matched fields
— is shared and unchanged across all forms.

## Performance

`partial_tweets` extracts 7 fields per tweet plus 2 per nested user, in document order. The
benchmark contains, in the same binary, a hand-tuned ordered `find_field` chain
(`simdjson_ondemand`) and the key-selector version (`simdjson_ondemand_key_selector`, written in
the mixed direct-binding + lambda style shown above), so the two are an A/B inside one process.

Measured with gcc 16.1.0 (best bytes/sec over repeated runs, `--benchmark_min_time` ≥ 0.4s):

| Platform | ordered `find_field` baseline | key selector | 
|---|---|---|
| Apple arm64 | ~6.75 GB/s | ~6.52 GB/s |
| x64 | ~6.65 GB/s | ~6.38 GB/s |

The key selector trades a small amount of throughput for order independence and a single pass.
Note that at this precision the figures are sensitive to binary code layout: the *unchanged*
ordered baseline alone moves by ~2% across builds whose only difference is the selector
participant's code, so the apparent gap should be read with that in mind. A version using
`error_code`-returning lambdas measures the same as the direct-binding version, i.e. per-field
error propagation is not a measurable cost here.

## Static-reflection deserialization — `doc.get<T>()`

The same `key_selector` / `for_each` machinery backs the static-reflection deserialization path
(`doc.get<T>()`), where extracting a struct's fields in one order-independent pass replaces a
chain of per-field ordered lookups.

  Full-struct deserialization of a reflected type. On `main` this issues one `obj[key]`
  lookup per member; this PR walks the object once with a compile-time perfect-hash
  `key_selector`. **Faster on both workloads and both architectures:**

  | Workload | Machine | main `obj[key]` | this PR `key_selector` | Speedup |
  |---|---|---:|---:|---:|
  | CITM (1.7 MB)    | x64 (Xeon Gold 6548N) | 2.07 GB/s | 2.54 GB/s | **+23.0 %** |
  | CITM (1.7 MB)    | ARM (Apple M4 Max)    | 3.08 GB/s | 3.73 GB/s | **+21.0 %** |
  | Twitter (631 KB) | x64 (Xeon Gold 6548N) | 4.39 GB/s | 4.81 GB/s | **+9.6 %**  |
  | Twitter (631 KB) | ARM (Apple M4 Max)    | 4.73 GB/s | 5.00 GB/s | **+5.7 %**  |

## Tests

`tests/ondemand/ondemand_object_find_field_tests.cpp`:

- Index form: void and `error_code` callbacks, error propagation, conditional-`noexcept`
  `static_assert`s, and (in exception builds) a throwing-callback propagation test.
- Per-key forms: direct variable binding (named selector and direct-key), `error_code` and void
  lambdas, mixed target + lambda, `simdjson_result<object>` forwarding, `N == 1`, absent keys,
  type-mismatch propagation, stop-on-error ordering, and nested-`for_each` result propagation.
- A dedicated `object_for_each_variadic_no_exceptions` test uses only non-throwing operations
  and runs in the `SIMDJSON_EXCEPTIONS=0` build, confirming the API works without exceptions.
- Matcher agreement, long-key, and `describe` tests for `key_selector` itself.

Verified on gcc 16.1.0 on both arm64 and x64, in three configurations: exceptions on (C++20 and
C++26/reflection) and **exceptions off** (`-DSIMDJSON_EXCEPTIONS=OFF`, `-fno-exceptions`,
C++20). All pass.

## Compatibility

- Additive: nothing outside the `SIMDJSON_SUPPORTS_CONCEPTS` blocks changes behavior.
- `for_each_result` gaining `[[nodiscard]]` is the only source-compatibility note: code that
  called `for_each` purely for its side effects and discarded the result must now assign it or
  cast to `void`.

## Files changed (high level)

- `include/simdjson/generic/ondemand/key_selector.h` — the compile-time selector and matcher.
- `include/simdjson/generic/ondemand/object.h`, `object-inl.h` — the `for_each` overloads, the
  `field_handler` concept, the dispatch adapter, `for_each_result`, and the forwarders.
- `benchmark/partial_tweets/simdjson_ondemand_key_selector.h` — the benchmark participant,
  written in the direct-binding + lambda style.
- `tests/ondemand/ondemand_object_find_field_tests.cpp` — the test coverage above.
- `doc/basics.md` — the "Key selectors" section and worked examples.

## Reproduction

```
# partial_tweets A/B (needs concepts, i.e. C++20+):
cmake -S . -B build -DSIMDJSON_DEVELOPER_MODE=ON -DSIMDJSON_CXX_STANDARD=20
cmake --build build --target bench_ondemand
./build/benchmark/bench_ondemand --benchmark_filter="partial_tweets<simdjson_ondemand"

# tests without exceptions:
cmake -S . -B build_noexc -DSIMDJSON_DEVELOPER_MODE=ON -DSIMDJSON_CXX_STANDARD=20 -DSIMDJSON_EXCEPTIONS=OFF
cmake --build build_noexc --target ondemand_object_find_field_tests
./build_noexc/tests/ondemand/ondemand_object_find_field_tests
```


lemire:
This is  marked as experimental, but pretty good now. I am merging.

PR #2608, opened 2026-02-20
Add LoongArch LSX support for find_next_json_quotable_character

Sudhanshu-S3:
#### Description
- Added a `#elif SIMDJSON_EXPERIMENTAL_HAS_LSX` path for
  `find_next_json_quotable_character` in `json_string_builder-inl.h`
  using LSX intrinsics, following the existing NEON/SSE2 pattern.
- Issue: #2603

#### Type of change
- [x] New feature

#### How to verify / test
Built locally on x86_64 — no regressions. The LSX path requires LoongArch hardware or CI to verify.

#### Checklist before submitting
- [ ] I added/updated tests covering my change (if applicable)
- [x] Code builds locally and passes my check
- [ ] Documentation / README updated if needed
- [x] Commits are atomic and messages are clear
- [x] I linked the related issue (if applicable)

PR #2778, opened 2026-07-08
Add LASX (LoongArch 256-bit SIMD) fast path for JSON string builder

vsaraikin:
### Summary

Adds LASX (LoongArch 256-bit SIMD) fast paths for the JSON string builder's escaping hot path, mirroring the existing SSE2 / NEON / LSX kernels:

- `fast_needs_escaping(std::string_view)` — 32-byte-at-a-time scan for `"`, `\`, and control chars (`< 0x20`).
- `find_next_json_quotable_character(view, location)` — 32-byte SIMD search that locates the first quotable byte via the four 64-bit lanes.

Both fall back to the scalar path for inputs shorter than one 256-bit vector, matching the LSX implementation's structure.

### Context

This continues the LoongArch SIMD work already merged:
- #2159 LSX/LASX support
- #2575 runtime dispatching for LoongArch
- #2608 LSX `find_next_json_quotable_character`
- #2727 LSX `fast_needs_escaping`

LSX (128-bit) covers both builder hot paths; LASX (256-bit) was still missing them, so on LASX-capable hardware these functions ran the narrower/scalar path. This PR closes that gap, guarded by `SIMDJSON_EXPERIMENTAL_HAS_LASX` (defined when `__loongarch_asx` is present) and `<lasxintrin.h>`.

### Notes

- No behavior change on any non-LoongArch target; the LASX branch is only selected when `__loongarch_asx` is defined.
- Draft: I don't have LASX hardware to benchmark on — I'd appreciate guidance on the project's preferred way to validate LoongArch SIMD paths (CI/QEMU), and will add any requested tests before marking ready.

### DIFF
--- a/include/simdjson/generic/builder/json_string_builder-inl.h
+++ b/include/simdjson/generic/builder/json_string_builder-inl.h
@@ -34,6 +34,11 @@
 #define SIMDJSON_EXPERIMENTAL_HAS_LSX 1
 #endif
 #endif
+#if defined(__loongarch_asx)
+#ifndef SIMDJSON_EXPERIMENTAL_HAS_LASX
+#define SIMDJSON_EXPERIMENTAL_HAS_LASX 1
+#endif
+#endif
 #if defined(__riscv_v_intrinsic) && __riscv_v_intrinsic >= 11000 &&            \
     defined(__riscv_vector)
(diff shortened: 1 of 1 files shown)


lemire:
Running tests.

Issue #2731, opened 2026-05-24
add u8 variant of string access methods


VanillaMaster:
Since all* strings in simdjson are guaranteed to be valid UTF-8,
it whould be handy to add "u8" variants of string access methods for C++20 users,
like `get_u8string`, `escaped_u8key` and so on.

AlbertoFVisconti:
Hi everyone, I would like to start contributing to this repo and this issue looks like a great starting point. If the maintainers are interested in this feature, I would gladly help!

lemire:
@AlbertoFVisconti It is not too difficult. When compiling with C++20, our `std::string_view` instances are effectively `std::u8string_view` instances. This could be added to our API. Should be only a few lines of code.

VanillaMaster:
@AlbertoFVisconti, @lemire technically speaking, it's somewhat tricky to make it "right" (without relying on UB), due to strict aliasing

lemire:
@VanillaMaster I don't think it is an issue in practice. A more important objection is that, as far as I can tell, nobody uses `char8_t`. So it is a bit of a marginal issue.

AlbertoFVisconti:
@lemire Since I'm using this as a learning experience, I went ahead and implemented the get_u8string API locally anyway. I guarded it with standard C++20 macros to ensure it didn't break C++17 builds, and wrote a test to verify it works on C++20.

If you want, I can open a PR just so you can take a look. If you think it adds too much clutter for a niche feature, feel free to close it.

lemire:
@AlbertoFVisconti Try it!

lemire:
Thanks to @AlbertoFVisconti, #2786 adds `get_u8string()` (returning `std::u8string_view`) to `dom::element` when `char8_t` is available (C++20). It's on `master` and will ship in the next release. Closing. If there's demand for the same thing in On Demand (e.g. `ondemand::value::get_u8string()`), please open a PR or a new issue.

PR #2786, opened 2026-07-21
Add C++20 u8 variants of string access methods (#2731)

AlbertoFVisconti:
Short title (summary):

Description
- What did you change and why? (1-3 sentences)
This PR adds C++20 `char8_t` support by introducing the `get_u8string()` method to `dom::element` and `simdjson_result<dom::element>`. This allows C++20 users to directly retrieve valid UTF-8 strings as `std::u8string_view` using a safe `reinterpret_cast` without needing to cast it manually. The implementations are strictly guarded by `#if defined(__cpp_char8_t) && __cpp_char8_t >= 201811L` to ensure zero impact on C++11/C++17 builds.

Type of change
- [ ] Bug fix
- [ ] Optimization
- [ x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

How to verify / test
- Added the `u8string_value()` test case to `tests/dom/basictests.cpp` to verify correct parsing and casting behavior. 
- Verified locally that building with `SIMDJSON_CXX_STANDARD=17` correctly ignores the macro guards.
- Verified locally that building with `SIMDJSON_CXX_STANDARD=20` compiles the new code, executes the test, and passes all assertions.

Checklist before submitting
- [x ] I added/updated tests covering my change (if applicable)
- [ x] Code builds locally and passes my check
- [x ] Commits are atomic and messages are clear
- [x ] I linked the related issue (if applicable)

### DIFF
--- a/include/simdjson/dom/element-inl.h
+++ b/include/simdjson/dom/element-inl.h
@@ -65,6 +65,12 @@ simdjson_inline simdjson_result<std::string_view> simdjson_result<dom::element>:
   if (error()) { return error(); }
   return first.get_string();
 }
+#if defined(__cpp_char8_t) && __cpp_char8_t >= 201811L
+simdjson_inline simdjson_result<std::u8string_view> simdjson_result<dom::element>::get_u8string() const noexcept {
+  if (error()) { return error(); }
+  return first.get_u8string();
+}
+#endif
 simdjson_inline simdjson_result<int64_t> simdjson_result<dom::element>::get_int64() const noexcept {
   if (error()) { return error(); }
   return first.get_int64();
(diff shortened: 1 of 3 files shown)


lemire:
Let us run tests!

AlbertoFVisconti:
I noticed the CI checks failed on the noexcept builds. I changed .value(), which throws an exception, for .value_unsafe() and now it should work!

include/simdjson/dom/element-inl.h
```diff
@@ -263,6 +269,15 @@ inline simdjson_result<std::string_view> element::get_string() const noexcept {
       return INCORRECT_TYPE;
   }
 }
+#if defined(__cpp_char8_t) && __cpp_char8_t >= 201811L
+inline simdjson_result<std::u8string_view> element::get_u8string() const noexcept {
+  simdjson_result<std::string_view> res = get_string();
+  if (res.error()) {
+      return res.error();
+  }
+  return std::u8string_view(reinterpret_cast<const char8_t*>(res.value_unsafe().data()), res.value_unsafe().size());
```
lemire: Try something like this...

```cpp
std::string_view v;
SIMDJSON_TRY(get_string().get(v));
```

I discourage the use of value_unsafe (as the name suggests, it is error prone).

AlbertoFVisconti: thanks for the feedback! I have updated it with the macro and pushed, i have tested locally with noexcept so it should work.

lemire:
@AlbertoFVisconti I have removed the claim that it removes  #2731 because it is only a partial step, for now.

PR #2795, opened 2026-07-28
Faster write_string_escaped: single-pass block escaping (SSE2, NEON)

lemire:
write_string_escaped scanned the input for the next quotable character and then copied the run, so every byte was visited twice, and the scan fell back to a byte-at-a-time table lookup for anything shorter than a SIMD block. Since the median JSON string is short (11 bytes in twitter.json, 8 in citm_catalog.json), most strings never reached the vector path at all.

Instead, consume the input in 16-byte blocks and copy a block with a single unaligned SIMD store when it holds nothing to escape, which makes one pass over the data. Blocks that do need escaping are fixed up from the escape mask, so whole runs between escapes are memcpy'd rather than rescanned.

The tail (fewer than 16 bytes) is covered by overlapping loads that stay entirely inside the input -- the last 16 bytes when the string is at least that long, otherwise two overlapping 8-byte or 4-byte loads. Nothing reads past the end of the string, so this is clean under sanitizers and safe for inputs ending at a page boundary. Strings shorter than 4 bytes stay scalar.

Measured against the two serialization benchmarks (best of three interleaved rounds):

  Xeon Gold 6548N, GCC 16     twitter        citm_catalog
    static_reflection          +17.8%          +4.8%
    reuse_buffer               +20.8%          +5.9%
    to                         +14.1%          +8.7%
    to_reuse                   +14.4%          +8.5%

I have not applied to systems other than x64 and ARM. This could be investigated in the future.

Reference https://github.com/simdjson/simdjson/discussions/2794

Credit to @fior512

### DIFF
--- a/include/simdjson/generic/builder/json_string_builder-inl.h
+++ b/include/simdjson/generic/builder/json_string_builder-inl.h
@@ -151,7 +151,11 @@ simdjson_inline bool fast_needs_escaping(std::string_view view) {
     running = vorrq_u8(running, vceqq_u8(word, v92));
     running = vorrq_u8(running, vcltq_u8(word, vdupq_n_u8(32)));
   }
-  return vmaxvq_u32(vreinterpretq_u32_u8(running)) != 0;
+  // `running` is an OR of comparison results, so every byte is 0x00 or 0xFF and
+  // the narrow-then-fcmp test applies: no across-lane reduction, no
+  // SIMD-to-general-purpose-register transfer. See escape_any below.
+  const uint8x8_t narrowed = vshrn_n_u16(vreinterpretq_u16_u8(running), 4);
+  return vdupd_lane_f64(vreinterpret_f64_u8(narrowed), 0) != 0.0;
 }
 #elif SIMDJSON_EXPERIMENTAL_HAS_SSE2
 simdjson_inline bool fast_needs_escaping(std::string_view view) {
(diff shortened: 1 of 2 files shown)


include/simdjson/generic/builder/json_string_builder-inl.h
```diff
@@ -564,6 +568,266 @@ SIMDJSON_CONSTEXPR_LAMBDA simdjson_inline void escape_json_char(char c, char *&o
...
+// The escape bitmask of a 16-byte block.
+simdjson_inline uint64_t escape_mask(escape_vector v) noexcept {
+  return escape_bitmask(escape_flags(v));
+}
+
+// Escapes the bytes of src in the range [i, blockend), given that m is the
+// (non-zero) escape mask for that range: the escape_mask_bits-wide lane of m
+// at byte k is non-zero when src[i + k] requires escaping. Returns the updated
+// output pointer.
+simdjson_inline char *escape_block(const uint8_t *src, char *out, size_t i,
```
fior512: can't we concat the head inside the loop directly ?
which will change the "meaning" of `pos` to "first byte not yet copied".

To make it look like:
```cpp
  constexpr uint64_t lane = (uint64_t(1) << escape_mask_bits) - 1;
  size_t pos = i; // first byte not yet copied
  while (m) { // avoid break structure
    const size_t tz = trailing_zeroes(m);
    const size_t next = i + tz / escape_mask_bits;
    // copy before escape to remove the -1/+1 offsets
    copy_lt16(out, src + pos, next - pos);
    out += next - pos;
    escape_json_char(char(src[next]), out);
    pos = next + 1; //step after escape
    m &= ~(lane << tz);
  }
  copy_lt16(out, src + pos, blockend - pos);
  return out + (blockend - pos);
```

fior512:
bench for the concatenated version proposed in review:
#### twitter
| Benchmark | PR MB/s | PR i/b | concat MB/s | concat i/b | MB/s | i/b |
|-----------|---------|--------|--------------|-------------|-------------|------------|
| static_reflection | 9772.85  | 2.35 | 10805.48 | 2.24 | +10.6% | -4.7% |
| reuse_buffer      | 12420.21 | 2.24 | 14234.30 | 2.13 | +14.6% | -4.9% |
| to                | 8855.51  | 2.40 | 9569.71  | 2.30 | +8.1%  | -4.2% |
| to_reuse          | 8811.14  | 2.41 | 9523.73  | 2.30 | +8.1%  | -4.6% |

#### citm_catalog
| Benchmark | PR MB/s | PR i/b | concat MB/s | concat i/b | MB/s | i/b |
|-----------|---------|--------|--------------|-------------|-------------|------------|
| static_reflection | 7236.35 | 3.23 | 7247.36 | 3.23 | +0.2% | 0.0%  |
| reuse_buffer      | 8374.10 | 3.24 | 8405.15 | 3.24 | +0.4% | 0.0%  |
| to                | 6614.24 | 3.26 | 6668.39 | 3.27 | +0.8% | +0.3% |
| to_reuse          | 6671.25 | 3.26 | 6699.61 | 3.27 | +0.4% | +0.3% |

> AMD Ryzen 5 7600X 6-Core Processor

---
PS: i deleted this comment accidentally earlier, sorry for the double notification.

include/simdjson/generic/builder/json_string_builder-inl.h
lemire: Sounds promising.

lemire:
@fior512 Verified.

PR #2799, opened 2026-07-30
perf: in-place to_json, integer-valued double fast path, SWAR integer parsing, in-register escape handling

FranciscoThiesen:
Three independent optimizations to the reflection serializer and on-demand integer parsing. Output is byte-identical to `master` in every benchmark and fuzz case; each change is motivated by a profile and ships with a dedicated correctness harness.

> **Rebased on current master** (#2795, #2798, #2804, #2805 all merged since the first revision; everything below is re-measured against b3072d25, 5 alternating co-built rounds, medians). Following the escaper comparison in the comments and the interest there, this PR now also carries a fourth change: the in-register escape handling grafted into #2795's `escape_block` structure. Details in section 4.

| benchmark (`benchmark/static_reflect`) | master | this PR |  |
|---|---:|---:|---|
| twitter serialize, reused `string_builder` | 13,757 MB/s | 14,696 MB/s | **+6.8%** |
| twitter `to_json(std::string&)`, reused string | 8,343 | 12,359 | **+48.1%** |
| twitter `to_json(std::string&)`, fresh string | 8,360 | 9,039 | +8.1% |
| citm `to_json(std::string&)`, reused string | 5,502 | 6,458 | +17.4% |
| citm deserialize (`doc.get<T>()`) | 3,352 | 3,551 | **+5.9%** |
| integer-valued doubles (micro) | 650 | 3,153 | **+385%** |
| escape-heavy strings, 1 escape / 16 B (micro) | 980-1,143 | 1,475-1,796 | **+32-59%** |

<sub>gcc 16.1, aarch64/NEON, `-O3`. Baseline and patched trees co-built in one container, 5 alternating rounds, medians (same-invocation A/B only; cross-run numbers drift on this hardware). twitter deserialize and the reused-`string_builder` serialization paths are unchanged (±noise).</sub>

---

### 1. `to_json(value, std::string&)` builds in place

Profile of `to_json` on twitter (taken before #2795; the copy share only grows with the faster escaper): ~43% of the time goes to copying or managing bytes that were already produced:

```
write_string_escaped  █████████████████░░░░░░░░░  41%
memcpy / growth       ███████████░░░░░░░░░░░░░░░  28%   ← this change
to_json wrapper       ██████░░░░░░░░░░░░░░░░░░░░  15%   ← this change
```

The old implementation builds in a private buffer (doubling from 1 KB, each doubling a `new[]`+`memcpy`), then copies the whole result *again* into the destination:

```cpp
// before: every output byte written ~3x
string_builder b(initial_capacity);
append(b, z);
s.assign(view);          // full copy of the output

// after: the destination's storage is the build buffer
string_builder b(s, initial_capacity);   // new string-backed mode
append(b, z);
s.resize(view.size());   // shrink, no copy; capacity kept for the next call
```

A caller that reuses the string pays **zero allocations and zero copies**, which is the common production pattern. `to_json_string` builds into a local and move-returns. The owned mode is untouched on its hot path (the two new members are cold; a cached `buf` pointer keeps codegen branch-free).

*Caveats:* destination must not alias the serialized value (documented); under `-fno-exceptions`, `std::string` growth failure aborts exactly as it would in caller code.

### 2. Integer-valued doubles take the integer writer

`double` fields holding exact integers (ids, counts, epoch timestamps) are everywhere in real JSON, and the shortest round-trip formatting of such a value below 10^15 is just its integer digits plus `".0"`:

```cpp
if (d > -1.0e15 && d < 1.0e15) {          // false for NaN, so the cast is safe
  int64_t iv = static_cast<int64_t>(d);
  if (static_cast<double>(iv) == d && !(iv == 0 && std::signbit(d))) {
    p = internal::write_uint_jeaiii(p, mag);   // ~5x faster than the float path
    *p++ = '.'; *p++ = '0';
```

Previously ~2.4× behind yyjson on this data class, now slightly ahead. Full-precision doubles pay 1–2% for the check.

The bound is 10^15, **not** 2^53: `to_chars` switches to scientific notation at 16 significant digits. An exhaustive identity sweep (2,403,929 cases: powers of 2/10, trailing-zero grids, the ±2^53 band, random integer-valued doubles) proves the fast path byte-identical to `to_chars` over the whole gate. The sweep caught an earlier 2^53 version of this patch and is worth keeping as a test.

*Bundled fix:* with `SIMDJSON_ENABLE_NAN_INF=0` (the default), `append(NaN)` used to reach `to_chars` and emit `2.696539702293474e+308`, valid JSON with silently wrong data. It now emits `null`. Happy to switch to returning an error if preferred.

### 3. SWAR fast path for `get_int64` / `get_uint64`

Profile of citm deserialize (baseline): `stage1` 40%, struct deserialization 32%, and inside the latter, digit-at-a-time integer parsing. The code declines SWAR with a comment: *“we don't use is_made_of_eight_digits_fast because large integers like 123456789 are rare.”* In id-heavy documents they are the norm:

```cpp
// gate = the validity check itself: one 8-byte load (in bounds via
// SIMDJSON_PADDING); short numbers fall through to the byte loop unchanged
if (simdjson_likely(!is_made_of_eight_digits_fast(src))) { return parse_unsigned(src); }
uint64_t i = parse_eight_digits_unrolled(src);           // 8 digits per step
if (is_made_of_eight_digits_fast(src + 8)) { i = i * 100000000 + ...; }
```

9/16/19-digit integers: +49% / +101% / +81% in isolation; citm deserialize +5.8%; twitter neutral. On targets that set `SIMDJSON_SWAR_NUMBER_PARSING` to 0 (big-endian, where `parse_eight_digits_unrolled` is not safe) both functions fall back to the byte loop. *Trade-off, disclosed:* documents of exclusively 2-5-digit integers measure −8 to −13% on the gate in a synthetic micro; both real-document benchmarks are net positive or neutral.

Correctness: a differential fuzz (15,528 adversarial cases: int64/uint64 boundaries ±1, 20-digit overflows, leading zeros, adjacent numbers straddling the 16-byte window, non-minified spacing) produces bit-identical (value, error) pairs against baseline.

---

### 4. Escape-heavy blocks are consumed in registers

Follow-up to the escaper comparison in the comments (and thanks @fior512 for the offer to help; this is the merge of the two approaches). #2795's structure is unchanged for clean blocks and sub-16-byte strings. When a block *does* contain escapes, it is now blind-stored and its whole mask is consumed in registers: after each escape expansion the remaining lanes are repositioned with a shuffle (`vqtbl1q_u8` on NEON, a one-time spill on SSE2) instead of `escape_block`'s per-run copies.

```cpp
escape_store16(out, word);                    // clean prefix lands in place
out = escape_block_hot(word, src, out, i, escape_bitmask(flags));
// hot: per escape, emit + reposition the tail from the register; one load
// and one store per 16 input bytes at any escape density
```

Escape-heavy strings gain +32-59% at every length (one escape per 16 bytes); twitter serialization gains ~6% across the builder variants. Disclosed trade-off: long strings with *sparse* escapes (one per ~333 bytes) measure 5-11% slower in a synthetic sweep, because the out-of-line hot call re-materializes the vector constants per escape-bearing block; documents like twitter/citm come out ahead (see table). I explored keeping the hot path inline plus a length split (four variants measured); each traded a different band. Happy to share that data in a follow-up if there is appetite for per-block strategy selection.

The blind store needs one vector of slack past the worst-case expansion, so the escape capacity checks move from `2 + 6*len` to `6*len + 18` (and analogues at all five call sites); `escape_tight_capacity_boundary` exercises exactly those budgets under ASan, alongside #2804's exhaustive escaper test which also passes.

**Testing.** The second commit adds `escape_tight_capacity_boundary` to `builder_string_builder_tests`: dense escape patterns at every block-boundary length/offset, built with `initial_capacity = 1` so every growth lands on an exact `pos+needed` boundary, byte-compared against a scalar reference. It guards the #2795 escaper's capacity handling and passes under ASan. The other harnesses (double-format identity sweep, number-parse differential fuzz) are standalone; glad to contribute them as unit tests too.

**Scope.** All numbers are gcc 16 / aarch64; x86 validation welcome. The three changes are independent, happy to split if easier to review.

### DIFF
--- a/include/simdjson/generic/builder/json_builder.h
+++ b/include/simdjson/generic/builder/json_builder.h
@@ -443,23 +443,30 @@ void append(string_builder &b, const Z &z) {
 }
 
 
-template <class Z>
-simdjson_warn_unused simdjson_result<std::string> to_json_string(const Z &z, size_t initial_capacity = string_builder::DEFAULT_INITIAL_CAPACITY) {
-  string_builder b(initial_capacity);
-  append(b, z);
-  std::string_view s;
-  if(auto e = b.view().get(s); e) { return e; }
-  return std::string(s);
-}
+// The convenience entry points below (to_json/to_json_string) build directly
+// into the destination std::string: its storage is the builder's buffer, so
+// there is no final copy, and a caller that reuses the same string across
+// calls pays no per-call allocation either (the string keeps its capacity).
+// Callers must not pass a destination that aliases the value being
+// serialized.
 
 template <class Z>
 simdjson_warn_unused error_code to_json(const Z &z, std::string &s, size_t initial_capacity = string_builder::DEFAULT_INITIAL_CAPACITY) {
-  string_builder b(initial_capacity);
+  string_builder b(s, initial_capacity);
   append(b, z);
   std::string_view view;
-  if(auto e = b.view().get(view); e) { return e; }
-  s.assign(view);
+  if(auto e = b.view().get(view); e) { s.clear(); return e; }
+  // The output was written in place; drop the unused tail (no copy, the
+  // string keeps its capacity for the next call).
+  s.resize(view.size());
   return SUCCESS;
+}
+
+template <class Z>
+simdjson_warn_unused simdjson_result<std::string> to_json_string(const Z &z, size_t initial_capacity = string_builder::DEFAULT_INITIAL_CAPACITY) {
+  std::string out;
+  if (auto e = to_json(z, out, initial_capacity); e) { return e; }
+  return out;
 }
 
 template <class Z>
@@ -523,21 +530,21 @@ simdjson_warn_unused simdjson_result<std::string> extract_from(const T &obj, siz
 } // namespace SIMDJSON_IMPLEMENTATION
 // Alias the function template to 'to' in the global namespace
 template <class Z>
-simdjson_warn_unused simdjson_result<std::string> to_json(const Z &z, size_t initial_capacity = SIMDJSON_IMPLEMENTATION::builder::string_builder::DEFAULT_INITIAL_CAPACITY) {
-  SIMDJSON_IMPLEMENTATION::builder::string_builder b(initial_capacity);
-  SIMDJSON_IMPLEMENTATION::builder::append(b, z);
-  std::string_view s;
-  if(auto e = b.view().get(s); e) { return e; }
-  return std::string(s);
-}
-template <class Z>
 simdjson_warn_unused error_code to_json(const Z &z, std::string &s, size_t initial_capacity = SIMDJSON_IMPLEMENTATION::builder::string_builder::DEFAULT_INITIAL_CAPACITY) {
-  SIMDJSON_IMPLEMENTATION::builder::string_builder b(initial_capacity);
+  // Build directly into `s` (its storage is the builder's buffer); see
+  // builder::to_json above.
+  SIMDJSON_IMPLEMENTATION::builder::string_builder b(s, initial_capacity);
   SIMDJSON_IMPLEMENTATION::builder::append(b, z);
   std::string_view view;
-  if(auto e = b.view().get(view); e) { return e; }
-  s.assign(view);
+  if(auto e = b.view().get(view); e) { s.clear(); return e; }
+  s.resize(view.size());
   return SUCCESS;
+}
+template <class Z>
+simdjson_warn_unused simdjson_result<std::string> to_json(const Z &z, size_t initial_capacity = SIMDJSON_IMPLEMENTATION::builder::string_builder::DEFAULT_INITIAL_CAPACITY) {
+  std::string out;
+  if (auto e = to_json(z, out, initial_capacity); e) { return e; }
+  return out;
 }
 // Global namespace function for extract_from
 template<constevalutil::fixed_string... FieldNames, typename T>
--- a/include/simdjson/generic/builder/json_builder.h
+++ b/include/simdjson/generic/builder/json_builder.h
@@ -151,8 +151,8 @@ simdjson_really_inline constexpr void atom(writer &w, const T &t) {
   // wrapped to a small value, ensure() would spuriously succeed and the
   // subsequent escape would overflow the buffer.
   // Note that this is pedantic except maybe on 32-bit targets.
-  if (simdjson_unlikely(input.size() > ((std::numeric_limits<size_t>::max)() - 2) / 6)) { return; }
-  if (!w.ensure(2 + 6 * input.size())) { return; }
+  if (simdjson_unlikely(input.size() > ((std::numeric_limits<size_t>::max)() - 18) / 6)) { return; }
+  if (!w.ensure(6 * input.size() + 18)) { return; }
   w.ptr[w.pos++] = '"';
   w.pos += write_string_escaped(input, w.ptr + w.pos);
   w.ptr[w.pos++] = '"';
@@ -182,8 +182,8 @@ simdjson_really_inline constexpr void atom(writer &w, const T &m) {
     // wrapped to a small value, ensure() would spuriously succeed and the
     // subsequent escape would overflow the buffer.
     // Note that this is pedantic except maybe on 32-bit targets.
-    if (simdjson_unlikely(key_sv.size() > ((std::numeric_limits<size_t>::max)() - 3) / 6)) { return; }
-    if (!w.ensure(2 + 6 * key_sv.size() + 1)) { return; }
+    if (simdjson_unlikely(key_sv.size() > ((std::numeric_limits<size_t>::max)() - 19) / 6)) { return; }
+    if (!w.ensure(6 * key_sv.size() + 19)) { return; }
     w.ptr[w.pos++] = '"';
     w.pos += write_string_escaped(key_sv, w.ptr + w.pos);
     w.ptr[w.pos++] = '"';
--- a/include/simdjson/generic/builder/json_string_builder-inl.h
+++ b/include/simdjson/generic/builder/json_string_builder-inl.h
@@ -667,9 +667,34 @@ inline size_t write_string_escaped(const std::string_view input, char *out) {
 
 
 simdjson_inline string_builder::string_builder(size_t initial_capacity)
-    : buffer(new(std::nothrow) char[initial_capacity]), position(0),
-      capacity(buffer.get() != nullptr ? initial_capacity : 0),
-      is_valid(buffer.get() != nullptr) {}
+    : buf(nullptr), position(0), capacity(0),
+      buffer(new(std::nothrow) char[initial_capacity]), is_valid(false) {
+  buf = buffer.get();
+  if (buf != nullptr) {
+    capacity = initial_capacity;
+    is_valid = true;
+  }
+}
+
+inline string_builder::string_builder(std::string &dest, size_t initial_capacity)
+    : buf(nullptr), position(0), capacity(0), buffer(), is_valid(true),
+      external(&dest) {
+  // Expose the string's full storage (at least initial_capacity) to the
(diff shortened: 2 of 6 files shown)


FranciscoThiesen:
Rebased on current master. #2795 merged a single-pass escaper while this PR was open — it is notably faster than the escaper rewrite this PR originally carried (especially on short strings), so that change is dropped and all numbers in the description are re-measured against #2795. Two artifacts from that work remain useful and are included/available:

- The second commit's `escape_tight_capacity_boundary` test now guards the **#2795** escaper: dense escape patterns at every block-boundary length/offset with `initial_capacity = 1` builders, byte-compared against a scalar reference; passes (incl. under ASan) on current master.
- In an isolated head-to-head, a blind-store escaper variant that repositions the chunk tail in registers measured +44–71% over `escape_block` at moderate escape densities (one escape per 16–64 bytes — e.g. JSON-in-JSON payloads), at parity on clean input and behind #2795 on short strings. If there is interest I can open a follow-up with the code and numbers for just that band.

fior512:
Hey @FranciscoThiesen, i am curious about the solution you had for escape chars before #2795; in #2794 i showed that performance gain decreases rapidly with strings >32 bytes, was your solution more reliable against long strings ?

lemire:
@FranciscoThiesen CI failures. ❤️  Got to love those.

FranciscoThiesen:
@lemire Fixed. Four independent causes, all ours:
1. `-Weffc++ -Werror`: the new raw `buf` member needed explicit copy-control on `string_builder` (copies now deleted, moves defaulted).
2. The new test used the concepts-gated `append(string_view)` overload, which doesn't exist in C++17 configurations (macOS/MinGW/fuzzing); it now uses `escape_and_append_with_quotes` directly.
3. `parse_*_swar` now fall back to the byte loop on targets with `SIMDJSON_SWAR_NUMBER_PARSING == 0`: `parse_eight_digits_unrolled` is not endian-safe there, so this one was a latent big-endian correctness bug (s390x, big-endian ppc64), not just a build issue.
4. The string-backed builder used `dest.data()` for mutable access, which is const-only before C++17 and broke every C++11/14 job (g++-13, clang-16, MSVC, UBSan). Now `&dest[0]`.

The full matrix is green as of the latest push.

FranciscoThiesen:
@fior512 Yes, and your >32 B instinct is almost exactly where the crossover sits. I ran a length sweep to check (gcc 16.1, aarch64, 5 alternating co-built rounds, medians; escapes placed randomly at 0.3% to match the twitter-like rate from your #2794 table):

| len (bytes) | #2795 | mine | Δ |
|---|---:|---:|---|
| 8 | 3.7 GB/s | 1.3 GB/s | **−64%** |
| 12 | 5.1 | 1.5 | **−71%** |
| 16 | 6.4 | 6.4 | ±0% |
| 24 | 6.0 | 3.7 | **−38%** |
| 32 | 9.1 | 9.4 | +4% |
| 48 | 10.4 | 11.4 | +10% |
| 64 | 11.6 | 13.1 | +13% |
| 128 | 13.2 | 16.0 | +21% |
| 256 | 13.7 | 17.8 | +29% |
| 512 | 13.7 | 18.3 | **+34%** |
| 2048 | 14.6 | 18.4 | +26% |
| 65536 | 14.5 | 18.0 | +24% |

The 16 B parity and the 24 B dip are the same effect seen from both sides: 16 B is exactly one SIMD chunk for my version, 24 B leaves an 8-byte scalar tail, and your overlapped loads handle both without a tail. On clean input (measured separately at rate 0) it's parity at 16 B, 32 B and from ~512 B up, while at 48-256 B clean #2795 is 2-6% faster. Below 16 B you win outright no matter what. At a denser 1/16 rate mine wins at every length, +22% to +81%.

One caveat on the numbers: the "mine" build also carried unrelated builder changes from my PR. A control run with identical escapers on both sides showed that tree loses 0-7% on this micro from binary layout alone, so the table, if anything, understates my side.

Why does length cancel out? Both escapers spend their steady state in the same loop, which compiles to 13 instructions per clean 16-byte chunk here (`ldr q`, `cmeq` x2 + `cmhs`, `orr` x2, `shrn` + `fcmp`, `b.ne`, `str q`, `add` + `cmp` + `b`). So roughly:

```
cycles/byte ≈ c_clean/16 + (k/n) · c_escape
```

n cancels, the density k/n survives, and c_escape dwarfs a clean chunk. The real difference between the two implementations is only what happens on a hit. `escape_block` copies the clean runs between escapes, so bytes that were already sitting in a register get copied again, and dense input turns into many small memcpys. Mine blind-stores the chunk, then consumes the whole mask in registers, repositioning the tail after each escape (`vqtbl1q_u8` on NEON, a one-time stack spill on SSE2): one load and one store per 16 input bytes at any density. That's the +13 to +34% band above. The price is shuffle setup on every hit, so it loses slightly at every-byte density and has no answer to your overlapped loads under 16 B, which is why I dropped it when #2795 landed.

Rereading #2794 with this data, I think your masked-load idea and mine are complements rather than competitors. Yours is strongest exactly where the per-string fixed costs live, 8-32 B, which covers the twitter/citm medians of 8-11 B and is why #2795 wins the macro benchmarks. The blind-store bulk path takes over from ~32 B up. A hybrid (#2795's short and clean paths unchanged, with in-register repositioning inside `escape_block`) should beat both everywhere. Happy to open a follow-up with the code and the sweep harness if there's interest.

lemire:
Promising.

fior512:
@FranciscoThiesen, sorry for the delay ^^. I am not completely sure i understand the idea you had about `write_string_escape`, but if you think both can be merged into a single one, i'd be glad to help !!

FranciscoThiesen:
@fior512 went ahead and merged the two approaches here (commit f5d4c4b2, the description now has fresh numbers against current master). Your structure stays exactly as is for clean blocks and short strings. When a block actually contains escapes, it now gets blind-stored and the whole mask is consumed in registers, instead of copying the runs between escapes.

Net effect: +32-59% on escape-heavy strings at any length, about +6% on twitter serialization, parity on clean input.

One trade-off I want to be upfront about: long strings with rare escapes (one per ~333 B) come out 5-11% slower in a synthetic sweep. The out-of-line escape handler makes the caller re-materialize its vector constants after each call. I tried four other shapes (inlining the handler, forcing the parent inline, a 32 B length split, a two-phase loop) and each one fixed that band by breaking another, so I kept the one that never loses on real documents.

If you feel like digging into per-block strategy selection together, I have all the sweep data and harnesses ready.

lemire:
Still on my todo.

FranciscoThiesen:
Rebased on current master to clear the conflict (one hunk in the number-parsing stubs, resolved as the union of the new `parse_float` stubs and the SWAR ones). Since master picked up perf-relevant work in the meantime (#2808, #2820, #2825, #2827), I re-ran the full suite against today's master. Fresh numbers, 5 alternating co-built rounds, medians:

| benchmark | master | this PR |  |
|---|---:|---:|---|
| twitter serialize, reused builder | 13,902 MB/s | 14,869 MB/s | +7.0% |
| twitter `to_json`, reused string | 8,464 | 12,513 | +47.8% |
| citm `to_json`, reused string | 5,526 | 6,464 | +17.0% |
| citm deserialize | 3,383 | 3,561 | +5.3% |
| integer-valued doubles (micro) | 653 | 3,149 | +382% |
| escape-heavy strings, 1/16 rate (micro) | | | +32-58% |

Everything within the usual bands of the pre-rebase measurements, so the rebase and the new master commits cost nothing. Correctness re-verified in the same run: escaper fuzz (20,866 cases, ASan, tight capacities), double-format identity sweep (2.4M cases), and the number-parse differential fuzz produces bit-identical results against the new master baseline, which also confirms #2820's semantics are untouched by the SWAR paths.

PR #2811, opened 2026-08-06
Add FreeBSD CI

### LINKED ISSUE #2810: FreeBSD coverage has been gone since Cirrus shut down
The FreeBSD job has not run since Cirrus CI shut down on 2026-06-01, and
.cirrus.yml was last touched in 2020-12-12 -- it still asks for
freebsd-13-0-snap.

Meanwhile .github/workflows has grown to 43 files covering aarch64, alpine,
emscripten, loongarch64, ppc64, riscv64 (five RVV configurations), s390x,
macOS, msys2 and a dozen Ubuntu and Visual Studio variants. None of them is
a BSD.

I ran the same three commands ppc64.yml uses, in a FreeBSD VM booted under
QEMU on a normal ubuntu-latest runner:

```
cmake -DCMAKE_BUILD_TYPE=Release -DSIMDJSON_DEVELOPER_MODE=ON -DSIMDJSON_COMPETITION=OFF -B build
cmake --build build -j=4
ctest --output-on-failure --test-dir build
```

https://github.com/neilpang/simdjson/actions/runs/31000616707

```
FreeBSD 14.3-RELEASE      clang 19.1.7   120/120 passed   82s   job 7m03s
FreeBSD 15.1-RELEASE-p1   clang 19.1.7   120/120 passed   96s   job 7m44s
```

No source changes were needed. cmake reports "Either git is unavailable or
else it is too old. We are disabling checkperf targets." in the VM, so the
checkperf targets are skipped there; the 120 tests are unaffected.

This also answers #771, where the Cirrus setup was called out as expensive
because of image building. There is no third-party service and no image to
build here: it is a stock GitHub-hosted runner booting a published VM image
under QEMU, so it costs the same as any other job in the matrix.

Would a PR adding this be welcome? I would add .github/workflows/freebsd.yml
shaped like ppc64.yml, and drop the dead .cirrus.yml in the same change --
happy to leave the latter alone if you would rather keep it.

Disclosure, per AI_USAGE_POLICY.md: I used an AI assistant to draft the
workflow and drive the verification runs. I have read and reviewed
everything here and can answer questions about any of it. The numbers above
are from the linked run, not from the model.

neilpang:
Closes #2810.

Adds `.github/workflows/freebsd.yml`. Same three commands as `ppc64.yml`, run
in a FreeBSD VM booted under QEMU on a normal ubuntu-latest runner.

https://github.com/neilpang/simdjson/actions/runs/31066180881

```
FreeBSD 14.3   clang 19.1.7   120/120 passed   6m40s
FreeBSD 15.1   clang 19.1.7   120/120 passed   8m34s
```

No source changes needed.

Two versions rather than one because both are supported upstream, and the old
`.cirrus.yml` covered three. Drop the 14.3 row if that is too much.

cmake reports "Either git is unavailable or else it is too old. We are
disabling checkperf targets." in the VM, so checkperf is skipped there. The 120
tests are unaffected.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -1,0 +1,32 @@
+name: FreeBSD
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        release: ['14.3', '15.1']
+    steps:
+      - uses: actions/checkout@v6
+      - uses: vmactions/freebsd-vm@v1
+        name: Test
+        with:
+          release: ${{ matrix.release }}
+          usesh: true
+          copyback: false
+          cache-after-prepare: true
+          prepare: |
+            pkg install -y cmake git bash
+          run: |
+            cmake -DCMAKE_BUILD_TYPE=Release -DSIMDJSON_DEVELOPER_MODE=ON -DSIMDJSON_COMPETITION=OFF -B build
+            cmake --build build -j=$(sysctl -n hw.ncpu)
+            ctest --output-on-failure --test-dir build


lemire:
Merging as it cannot break the lib.

PR #2813, opened 2026-08-06
fixing a typo in if-logic of perfdiff.cpp

fior512:
`benchmark/dom/perfdiff.cpp` had an inverted angle bracket inside an if statement, causing the program to only be able to say "Performance Degradation" or "no obvious performance difference" and never a positive answer.

PS: Metric used is GB/s, meaning higher-is-better


lemire:
Merging eagerly as this cannot break the lib.

PR #2818, opened 2026-08-10
fix: LTO workaround

fior512:
### Summary

This PR adds a guard, when MSVC builds the shared library, INTERPROCEDURAL_OPTIMIZATION (IPO) is forced off on the simdjson target and a warning is emitted. The user's IPO setting is left alone; so MSVC static builds, other targets, and every other toolchain still get LTO.

The issue (#451) was MSVC's `/GL` (LTO) optimization produces non-COFF objects that bindexplib (used by `WINDOWS_EXPORT_ALL_SYMBOLS`) can't read, causing empty export tables and link errors (only affecting DLL).

### What This PR does not
1) It does not solve #451 (MSVC + DLL + LTO), but only make it usable everywhere else
2) It doesn't add/update any CI

### Bench
Serialization benchmark with twitter.json 
| Benchmark | noLTO (MB/s) | LTO (MB/s) | % Change |
|-----------|--------------|-----------|----------|
| static_reflection | 10,965.87 | 11,707.39 | +6.76% |
| reuse_buffer | 14,246.01 | 15,452.13 | +8.47% |
| to | 9,657.98 | 10,123.93 | +4.82% |
| to_reuse | 9,697.22 | 10,060.99 | +3.75% |
| reflect_cpp | 1,698.18 | 1,723.98 | +1.52% |
| Binary Size (bloaty) | 1.41 Mi | 819 Ki | -41.91% |

> CPU: AMD Ryzen 5 7600X (6cores)
> Governor: Performance
> Boost: ON

---
I made the documentation lightweight for users, happy to expand it based on your feedback !!

Furthermore, I could only test this on g++ and clang, release developer-mode build with `-D CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` compiles and passes tests. 

Two things need eyes on Windows (i was not able to test myself):
 - the guard itself, i.e.: `cmake -B build -D BUILD_SHARED_LIBS=ON -D CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` now produces a working DLL instead of failing;
 - `BUILD_SHARED_LIBS=ON` + `SIMDJSON_BUILD_STATIC_LIB=ON` scenario, where simdjson loses LTO but simdjson_static keeps it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,6 +265,24 @@ if(WIN32 AND BUILD_SHARED_LIBS)
   )
 endif()
 
+
+# MSVC doesn't support shared-lib + LTO.
+# This is a workaround to make sure both are never used together
+if(SIMDJSON_WINDOWS_DLL) # MSVC + Shared; defined in developer-options.cmake
+  if(CMAKE_INTERPROCEDURAL_OPTIMIZATION
+      OR CMAKE_INTERPROCEDURAL_OPTIMIZATION_DEBUG
+      OR CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE
+      OR CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO
+      OR CMAKE_INTERPROCEDURAL_OPTIMIZATION_MINSIZEREL)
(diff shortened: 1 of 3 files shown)


lemire:
Merged. Fixing the Visual Studio issue is nice.

PR #2823, opened 2026-08-14
docs: fix cmake -B build directory in HACKING.md

ezralicodes:
## Summary

Fixes broken CMake quickstart commands in `HACKING.md`.

`-B` requires a build-directory argument. The previous examples used `cmake -B -D …`, so CMake treated `-D` as the build directory name and the documented flow could not work.

## Change

Use the standard out-of-tree form:

```bash
cmake -B build -D SIMDJSON_DEVELOPER_MODE=ON .
cmake --build build
ctest --test-dir build
```

Same correction for the sanitizer quickstart block.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -23,7 +23,7 @@ Build Quickstart
 For non-Windows system,
 
 ```bash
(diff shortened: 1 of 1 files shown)


lemire:
thanks !!!

PR #2829, opened 2026-08-18
Add get_float to the number accessor benchmark

lemire:
Follow-up to #2807: add `ondemand_get_float()` to the number accessor benchmark.

It uses the same floating-point dataset as `get_double()` (5 / 12 / 19 digits) so the two parsers can be compared on identical input. DOM has no `get_float()`, so only the On Demand path is measured.

A dedicated target is available, matching the existing flags:

- `BENCH_ONLY_FLOAT` through `bench_number_accessors_float`

Also drops the unused `<endian.h>` include so the benchmark compiles on macOS.

Sample (`bench_number_accessors_float` on Apple Silicon):

```txt
Trial 1:
  5 digits:
    ondemand_get_float           :     9.16 ns/value     1.26 GB/s
  12 digits:
    ondemand_get_float           :    10.05 ns/value     1.84 GB/s
  19 digits:
    ondemand_get_float           :    11.37 ns/value     2.24 GB/s
```


fior512:
@lemire Thanks for finishing this up !! Sorry, I missed a function, I know this isn't the first time (PR about inlining mem* functions in to_chars()). If it's not too much hassle for you, feel free to drop a quick comment on my PRs next time you observe something incoherent. I'd like to get more practice in team-based development and OSS projects.

lemire:
@fior512 I don't merge incoherent pull requests. If I thought your work was bad, I would not merge it.

When you wrote your PR, we did not have yet float support. You could not have done better.

fior512:
@lemire understood. I didn't realize `get_float()` wasn't there yet. Thanks for clarifying !

PR #2832, opened 2026-08-20
SWAR key comparison in dom object lookup

vpelikh:
## Summary

The DOM `object::iterator::key_equals` used libc `memcmp`, and `key_equals_case_insensitive` used `simdjson_strncasecmp` (libc `strncasecmp` / `_strnicmp`). simdjson's own comment on `key_equals_case_insensitive` already notes "it might be worth rolling our own strncasecmp function, with vectorization."

This PR replaces both with a byte-level compare that **accumulates mismatches over a countable number of 8-byte chunks with no early return** — a loop shape the compiler widens to the native SIMD register width:

```cpp
const size_t nchunks = n / 8;
uint64_t mismatch = 0;
for (size_t c = 0; c < nchunks; c++) {
  uint64_t x = *(const uint64_t *)(a + c * 8);
  uint64_t y = *(const uint64_t *)(b + c * 8);
  mismatch |= x ^ y;              // (or the folded chunks, case-insensitive)
}
if (mismatch != 0) return false;
```

## What changed

- `key_equals` accumulates `x ^ y` over all whole 8-byte chunks, then a byte tail. Any nonzero accumulated XOR is a mismatch. The exact-match path is unchanged in behavior.
- `key_equals_case_insensitive` runs the same loop on each chunk's ASCII-lowercase fold: each byte in `'A'..'Z'` maps to `| 0x20`, everything else passes through, using a carry-free per-lane SWAR range test. This matches `strncasecmp` semantics exactly — two bytes match case-insensitively iff they are equal, or differ only by the ASCII case bit while both folded values are in `'a'..'z'`; non-ASCII bytes compare exactly and are never folded.

Behavior is unchanged in both methods; only the loop shape differs so the compiler can widen it to SIMD.

## Before vs. after

| Benchmark | before (memcmp/strncasecmp) | after (this PR) |
|---|---|---|
| `lookup_flat_array_keys` (exact) | 139 ns | 138 ns |
| `lookup_case_insensitive_keys` | 122 ns | 141 ns |

## Type of change

- [ ] Refactor / cleanup
- [ ] Bug fix
- [x] Optimization
- [ ] New feature
- [ ] Documentation / tests
- [ ] Other:

## Tests

- New regression test `dom_api_tests::object_key_matching` in `tests/dom/basictests.cpp` covering exact and case insensitive lookups, keys filling/exceeding an 8-byte chunk, non-ASCII keys, and the `"  abcdef"` / `" !abcdef"` pair that an earlier `has_zero` SWAR trick got wrong.
- Full suite passes (`ctest`).

```bash
cmake -B build -D SIMDJSON_SANITIZE=ON -D SIMDJSON_DEVELOPER_MODE=ON
cmake --build build
ctest --test-dir build
cmake --build build --target bench_dom_api
./build/benchmark/bench_dom_api --benchmark_filter='lookup_.*_keys'
```

## Related commits on this branch

- `99579e6b` SIMD-friendly key comparison in dom object lookup
- `045f7165` Use vectorizable accumulate loop for DOM key comparison
- `dc0bb5a4` Remove redundant minus check in visit_primitive (unrelated; can be dropped if
  you'd rather this PR contain only the key-comparison optimization)

### DIFF
--- a/include/simdjson/dom/object-inl.h
+++ b/include/simdjson/dom/object-inl.h
@@ -344,41 +344,62 @@ inline element object::iterator::value() const noexcept {
 
 namespace {
 
-// Key comparisons used by object::iterator::key_equals. They process the key 8 bytes per
-// uint64_t chunk (mirroring the byte-level key comparison in the Java DOM port), so the
-// string-buffer bytes are each loaded once, which helps the compiler keep the work in
-// registers and, for key_equals, vectorize the chunk equality. simdjson buffers are
-// padded, so reading an 8-byte chunk is safe as long as i + 8 <= n.
-
-// Has-zero high-bit mask: each byte lane's high bit is set iff that byte of x is zero.
-simdjson_inline uint64_t swar_has_zero(uint64_t x) {
-  return (x - 0x0101010101010101ULL) & ~x & 0x8080808080808080ULL;
-}
-
-// True if all 8 bytes of the chunk match (vectorized-equality-friendly chunk compare).
-simdjson_inline bool swar_chunk_equal(uint64_t a, uint64_t b) {
-  return swar_has_zero(a ^ b) == 0x8080808080808080ULL;
-}
-
 // ASCII-letter case folding used by key_equals_case_insensitive: two bytes match
 // case-insensitively iff they are equal, or differ only by the ASCII case bit (| 0x20)
 // while both folded values are in 'a'..'z'. Non-ASCII bytes compare exactly.
-simdjson_inline bool swar_byte_equal_nocase(uint8_t a, uint8_t b) {
+// This mirrors the behavior of strncasecmp for single-byte input.
+simdjson_inline bool ascii_byte_equal_nocase(uint8_t a, uint8_t b) {
   if (a == b) { return true; }
   uint8_t fa = (uint8_t)(a | 0x20);
   uint8_t fb = (uint8_t)(b | 0x20);
   return fa == fb && (fa >= (uint8_t)'a' && fa <= (uint8_t)'z');
 }
 
-simdjson_inline bool swar_bytes_equal_nocase(const uint8_t *a, const uint8_t *b, size_t n) {
-  size_t i = 0;
-  for (; i + 8 <= n; i += 8) {
-    for (size_t k = 0; k < 8; k++) {
-      if (!swar_byte_equal_nocase(a[i + k], b[i + k])) { return false; }
-    }
+// ASCII-letter case fold applied per byte to a full 8-byte chunk: each byte in
+// 'A'..'Z' (0x41..0x5a) is mapped to its lowercase form (| 0x20); every other byte
+// is passed through unchanged. Lanes are independent: the SWAR range tests below
+// detect 'A'..'Z' by comparing against low 7-bit values so no carry / borrow can
+// cross a byte boundary, and the result is a plain memcmp-comparable chunk.
+simdjson_inline uint64_t ascii_fold_lower_chunk(uint64_t x) {
+  const uint64_t high  = 0x8080808080808080ULL;
+  const uint64_t lower = 0x2020202020202020ULL;
+  // Standard carry-free "greater than n" per byte (Hacker's Delight style): the high
+  // bit is set in exactly the lanes whose low 7-bit value is > n. Each add is contained
+  // within its byte lane (we add at most 0x7f to at most 0x7f, then mask the high bit),
+  // so no carry can cross a byte boundary. This is only meaningful for bytes < 0x80,
+  // which is all we need: 'A'..'Z' live in 0x41..0x5a.
+  auto greater_than = [](uint64_t v, uint64_t n) -> uint64_t {
+    const uint64_t ones = 0x0101010101010101ULL;
+    const uint64_t low7 = 0x7f7f7f7f7f7f7f7fULL;
+    return (((v & low7) + ((0x7f7f7f7f7f7f7f7fULL - n * ones) | 1)) ^ v) & 0x8080808080808080ULL;
+  };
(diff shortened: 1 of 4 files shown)


include/simdjson/dom/object-inl.h
```diff
@@ -342,26 +342,82 @@ inline element object::iterator::value() const noexcept {
...
+// registers and, for key_equals, vectorize the chunk equality. simdjson buffers are
+// padded, so reading an 8-byte chunk is safe as long as i + 8 <= n.
+
+// Has-zero high-bit mask: each byte lane's high bit is set iff that byte of x is zero.
+simdjson_inline uint64_t swar_has_zero(uint64_t x) {
+  return (x - 0x0101010101010101ULL) & ~x & 0x8080808080808080ULL;
+}
+
+// True if all 8 bytes of the chunk match (vectorized-equality-friendly chunk compare).
+simdjson_inline bool swar_chunk_equal(uint64_t a, uint64_t b) {
```
lemire: Yeah, I think that's likely buggy. Note that we cannot assume that we have ASCII inputs.

vpelikh: Resolved by [045f716](https://github.com/simdjson/simdjson/pull/2832/commits/045f7165f62aa457f7082c5dc6f546d3189792ed).

vpelikh:
## Why "accumulate, no early return"

A plain early-exit loop (`if (x != y) return false;`) does **not** vectorize: the early return makes the trip count hard to bound and the accesses potentially faulting, so it stays scalar. The compiler reports:

```
Cannot vectorize potentially faulting early exit loop
Cannot vectorize uncountable loop
```

Restructuring to accumulate mismatches into a `uint64_t` and test once after the loop makes it countable and side-effect-free, which is what enables real SIMD. Confirmed with `-Rpass=loop-vectorize` on the actual build:

```
vectorized loop (vectorization width: 2, interleaved count: 4)
```

and the emitted object contains NEON `.16b` ops (arm64).

## What I tried before settling on the final version

1. **Early-exit scalar 8-byte chunks** (the obvious form). Fast on short keys, but not vectorized: clang rejects the loop for the reasons above.
2. **Hybrid with a `len >= 32` threshold**: scalar early-exit for short keys, accumulate loop for long keys. Proved **slower** than either extreme on the DOM benchmark: the size branch plus duplicated loop costs more than the (never-taken, for short keys) SIMD path gains. A threshold sweep (8/16/32/64) confirmed the branch overhead dominates for real JSON key lengths.
3. **Always-accumulate (final)**: same vectorizable loop at every length. This is what the PR ships; it removes the threshold entirely.

## Benchmark methodology and results

- Machine: Apple Silicon (arm64), `bench_dom_api --benchmark_filter='lookup_.*_keys'`, several interleaved runs of 9 repetitions; the table below is the median.
- The DOM benchmark object has short keys (well under the former 32-byte threshold), so this emphasizes the common short-key case.

| Benchmark | before (memcmp/strncasecmp) | final (always-accumulate) | delta |
|---|---|---|---|
| `lookup_flat_array_keys` (exact) | 139 ns | 138 ns | parity -- and now really SIMD |
| `lookup_case_insensitive_keys` | 122 ns | 141 ns | **-15% (regressed)** |

### Interpretation

- **Exact** path lands at parity with the old `memcmp` while genuinely vectorizing: a free win, and the clean part of the change.
- **Case-insensitive** path regresses ~15% on short-key workloads. The intermediate hybrid (scalar for `len < 32`, accumulate for `len >= 32`) also regressed to ~132 ns solely from the branch cost. On long keys (`>= 64` bytes) the vectorized accumulate loop clearly wins, but real JSON keys are short, so the net effect on this workload is a small regression.

### A length sweep (matching buffers, single call, ns)

Small standalone microbenchmark toggling each strategy and length in isolation (not part of the branch). Each column is a distinct function; lower is better.

- **exact early-exit** -- case-sensitive (`memcmp`-style) compare that stops at the first mismatching 8-byte chunk (`if (x != y) return false;`); not vectorized.
- **exact accumulate** -- case-sensitive compare that ORs `x ^ y` over all 8-byte chunks with no early return, then tests once. This is the vectorized form this PR ships (for the exact path it becomes the plain chunk loop).
- **nocase early-exit** -- case-insensitive compare (ASCII fold) that stops at the first mismatching folded 8-byte chunk; not vectorized.
- **nocase accumulate** -- case-insensitive compare (ASCII fold) that accumulates the folded-chunk XOR with no early return. This is the vectorized form this PR ships (for the case-insensitive path).

| len | exact early-exit | exact accumulate | nocase early-exit | nocase accumulate |
|---|---|---|---|---|
| 8   | 1.8  | 1.5  | 2.0 | 2.4 |
| 16  | 2.7  | 1.8  | 3.4 | 3.7 |
| 32  | 3.4  | 3.0  | 6.1 | 6.6 |
| 64  | 4.5  | 2.3  | 11.3 | 7.5 |
| 128 | 6.9  | 3.1  | 21.5 | 13.4 |
| 256 | 12.9 | 4.9  | 42.0 | 26.4 |

The accumulate form is at least as fast as early-exit for the exact path at every measured length, and clearly wins for long keys (both paths); the case-insensitive accumulate only pulls ahead from ~64 bytes.

## Open question for discussion

The honest trade-off: this makes the code genuinely vectorizable (honoring the "roll our own strncasecmp, with vectorization" comment, and the exact path now emits real SIMD), but on the representative short-key DOM benchmark the case-insensitive path costs ~15%. Options to discuss:

- Keep the always-accumulate version as shipped (simplest, exact path free).
- Reintroduce a tuned short-key fast path **only** for case-insensitive at a measured (low) threshold if the 15% matters on the nocase path.

FYI @lemire, I'd appreciate your take on which direction to go.

lemire:
I am not convinced.

I am going to close this PR for now. Please consider issuing a new PR if you believe that you hold a promising direction.

PR #2846, opened 2026-08-27
Add object::get_current_position()/revert_position() to avoid full rescans on optional-field misses

### LINKED ISSUE #1952: Reverting iter to previous position
In the case of using a one-pass version of the parser (using "find_fileld"), if the field is not found, then the object has to be reset, which is logical, but not for the case when we expect that there may not be a value with the passed keys. This results in performance degradation as we reprocess the keys. This can be avoided by adding the ability to remember the position of the iterator.

    Result result;

    auto error = object.find_field(key).get(result);

    switch (error) {
        case simdjson::error_code::SUCCESS:
            break;

        case simdjson::error_code::NO_SUCH_FIELD:
            object.reset();
            break;

        default:
            ...
    }

Make something like this

    auto position = object.get_current_position();
    
    Result result;
    
    auto error = object.find_field(key).get(result);
    
    switch (error) {
        case simdjson::error_code::SUCCESS:
            break;
    
        case simdjson::error_code::NO_SUCH_FIELD:
            object.revert_position(position);
            break;
    
        default:
            ...
    }

taimoor789:
**Problem**

Ordered `find_field()` only recovers from a `NO_SUCH_FIELD` miss via `reset()`, which rescans the object from the start even when most fields were already consumed. Closes #1952.

**Root cause**

There was no way to save/restore an intermediate scanning position, so probing for an optional field that wasn't next always paid a full rescan.

**What changed**

- `object::get_current_position()` / `revert_position()` (+ `simdjson_result<object>` forwarders), backed by a new opaque `object_position` (private fields, `friend class object`, so it can't be hand-built with an invalid position/depth).
- `value_iterator::reenter_at(token_position, depth_t)`: a new primitive that moves the iterator directly to a given position/depth.
- Benchmark comparing `reset()`, `revert_position()`, and `find_field_unordered()` on a miss.
- Tests: scalar-field, nested-object, and unconsumed-compound-value revert cases; a `simdjson_result<object>` wrapper-path test; an `OUT_OF_ORDER_ITERATION` guard test; a compile-fail test proving `object_position` can't be constructed outside `object`.
- `doc/basics.md`: usage example and invalidation rules.

**Why this approach**

The existing `json_iterator::reenter_child()` looked reusable but assumes the live depth is exactly one level shallower than the target and validates against per-depth bookkeeping that's only populated at a container's own start — neither holds for an arbitrary captured position (found via `SIMDJSON_ASSUME` failures in a `SIMDJSON_DEVELOPMENT_CHECKS` build while testing). `reenter_at()` instead mirrors the existing unconditional `move_at_start()`/`move_at_container_start()` pattern already used by `reset()`, asserting only that the position/depth values themselves are sane.

**Tests performed**

- Full `ctest` suite (122 tests): 100% pass in Debug+`SIMDJSON_DEVELOPMENT_CHECKS=ON` and in Release.
- Fail-without-fix check: removing the implementation produces a hard compile failure in the new tests; restoring it passes again.
- No new compiler warnings.
- Benchmark (3 trials, miss probed after 10/20 fields): `revert_position()` ~3.1-3.8 ns/char vs `reset()` ~4.3-4.7 ns/char on a miss.

**Limitations**

- Tested locally on macOS (Apple clang, arm64) only — not run against the project's CI matrix (other compilers/architectures).
- The `OUT_OF_ORDER_ITERATION` guard and `reenter_at()`'s assertions are active only in `SIMDJSON_DEVELOPMENT_CHECKS` (Debug) builds, consistent with existing conventions.


benchmark/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ if (TARGET benchmark::benchmark)
   target_compile_definitions(bench_number_accessors_float PRIVATE BENCH_ONLY_FLOAT)
   add_executable(bench_number_accessors_number bench_number_accessors.cpp)
   target_compile_definitions(bench_number_accessors_number PRIVATE BENCH_ONLY_NUMBER)
+  add_executable(bench_object_revert_position bench_object_revert_position.cpp)
```
lemire: the benchmark is guarded by `if (TARGET benchmark::benchmark)`, is that correct?

taimoor789: @lemire This benchmark doesn't use Google Benchmark at all, only counters (already linked unconditionally) and from/benchmark_helpers.h. It shouldn't have been inside that guard, I moved it out alongside benchfeatures/get_corpus_benchmark, in the latest commit.

lemire:
Running tests.

taimoor789:
The 3 issues are all the same issue from GitHub while CMake was fetching third-party dependencies during configure, before any of this PR's code compiled. Every other matrix leg passed. Could you rerun the failed jobs?

lemire:
@taimoor789 Don't worry about that. I do restart failed CI tests.

taimoor789:
Perfect, thanks!

lemire:
Merged.

Issue #2806, opened 2026-08-01
CI Fuzz fail due to curl `connection failed`


fior512:
Fuzz/valgrind CI is failing at the "Store the corpus externally" step, because curl is unable to reach target.

observed for the first time in #2805; i can't reach target either (manually), likely server-down.
```
curl: (7) Failed to connect to www.pauldreik.se port 443 after 36940 ms: Couldn't connect to server
Error: Process completed with exit code 7.
```

fior512:
server is back online.

PR #2847, opened 2026-08-27
ci: don't let the external corpus upload abort the fuzz job

advitrocks9:
`Store the corpus externally` in `fuzzers.yml` is a bare `curl` to `www.pauldreik.se` with no
failure handling, and it runs before the valgrind replay. When the host is unreachable the job
stops there and the valgrind replay and the corpus cache save are skipped.

Of the last 1000 runs on master, 14 failed; 10 of the 11 with step data failed at this step.
#2806 was the same outage.

`continue-on-error: true` lets the rest of the job finish. The step still shows as failed in the
run, and the corpus is already uploaded as a GitHub artifact one step earlier, so nothing is lost
while the mirror is down.

Type of change
- [ ] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests
- [x] Other (please describe): CI reliability

Checklist before submitting
- [ ] I added/updated tests covering my change (if applicable)
- [ ] Code builds locally and passes my check
- [ ] I linked the related issue (if applicable)

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -131,6 +131,7 @@ jobs:
         path: corpus.tar
 
     - name: Store the corpus externally
+      continue-on-error: true
       run: |
         gzip --keep corpus.tar
         curl -F"filedata=@corpus.tar.gz" https://simdjson:${{ secrets.fuzzdatapassword }}@www.pauldreik.se/fuzzdata/index.php


lemire:
Ok, but what is the point of this change ?

advitrocks9:
It just allows the failed jobs to be resumed

advitrocks9:
If the corpus upload fails the whole fuzz job goes red, so an infra flake looks the same as a real fuzz failure.

PR #2855, opened 2026-09-05
Fix DOM at_path error propagation

lilong555:
This PR proposes preserving the original error when `at_path()` is called on a failed DOM result, consistent with the existing behavior of `at_pointer()` and the On-Demand `at_path()` wrappers.

For example, after parsing `{"array": [1], "object": {"key": 2}}`, `doc["missing"].at_path("invalid")` currently returns `INVALID_JSON_POINTER`, although the field lookup has already failed with `NO_SUCH_FIELD`. Checking the stored error before converting the path allows the original error to reach the caller.

The change adds an early error check to the `dom::element`, `dom::array`, and `dom::object` result wrappers. Regression tests cover missing fields, incorrect types, out-of-bounds indices, and parse failures, as well as path validation and successful queries. They also run with exceptions disabled.

Validation on Windows x64 with MSVC 19.44:

- The new regression test fails against the unmodified library with `INVALID_JSON_POINTER` instead of `NO_SUCH_FIELD`, and passes with the fix.
- Debug build: `ctest --test-dir build-contrib -C Debug -LE explicitonly --output-on-failure` — 112 tests passed.
- Release build with `SIMDJSON_EXCEPTIONS=OFF`: `ctest --test-dir build-contrib-noexcept -C Release -L dom --output-on-failure` — 21 tests passed.
- `git diff --check` and an ASCII check of all changed source files passed.

AI assistance was used to identify the issue and prepare the implementation, regression tests, and PR text.

Thank you for taking a look. Feedback on the error-handling behavior or test coverage would be very welcome.

### DIFF
--- a/include/simdjson/dom/array-inl.h
+++ b/include/simdjson/dom/array-inl.h
@@ -47,6 +47,7 @@ inline simdjson_result<dom::element> simdjson_result<dom::array>::at_pointer(std
 }
 
  inline simdjson_result<dom::element> simdjson_result<dom::array>::at_path(std::string_view json_path) const noexcept {
+  if (error()) { return error(); }
   auto json_pointer = json_path_to_pointer_conversion(json_path);
   if (json_pointer == "-1") { return INVALID_JSON_POINTER; }
   return at_pointer(json_pointer);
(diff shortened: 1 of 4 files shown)


lemire:
Sounds reasonable. Running tests.

lemire:
Looks non critical, but nice to have, merging to main.

PR #2862, opened 2026-09-17
Remove unused competition dependencies (cjson, jsmn, gason, fastjson,…

### LINKED ISSUE #2859: Several competition libraries are built but never benchmarked
`dependencies/CMakeLists.txt` fetches and builds cjson, jsmn, gason, fastjson, ujson4c and jsoncpp, defining a `SIMDJSON_COMPETITION_*` macro for each. Those macros appear nowhere else in the repository, so no benchmark uses them:

```
$ grep -rln "SIMDJSON_COMPETITION_CJSON\|SIMDJSON_COMPETITION_JSMN\|SIMDJSON_COMPETITION_GASON\|SIMDJSON_COMPETITION_FASTJSON\|SIMDJSON_COMPETITION_UJSON4C\|SIMDJSON_COMPETITION_JSONCPP" .
./dependencies/CMakeLists.txt
```

cjson and jsmn are also linked into `competition-core`, so they are compiled into the benchmark binary regardless.

This resembles #1391 for Boost.JSON, which was resolved by adding benchmarks in #2092. Would you prefer these added to the existing benchmarks following the boostjson/yyjson pattern, or the unused dependencies dropped? I am happy to do either.

pratap9572singh:
dependencies/CMakeLists.txt fetched and built cjson, jsmn, gason, fastjson, ujson4c and jsoncpp, none of which are used under benchmark/. Their only consumer was tests/dom/allparserscheckfile.cpp, which compared simdjson against rapidjson and sajson alongside those six; after this change it still compares rapidjson and sajson.

While removing them I found that target is never built. The guard at tests/dom/CMakeLists.txt:53 uses (!SIMDJSON_SANITIZE) rather than (NOT SIMDJSON_SANITIZE); with SIMDJSON_SANITIZE off the target is still not generated, and `cmake --build build --target allparserscheckfile` reports no such target. This dates to 89b059b1 (June 2020), and also explains the undeclared `err` at line 96 — the file does not currently compile. This PR fixes that identifier (every other reference in the function uses `error`, declared at line 42) but leaves the guard alone; happy to fix it separately if you'd like the checker building again.

Also removes the vendored dependencies/jsoncppdist/. competition-core is now nlohmann_json + rapidjson + sajson + yyjson; competition-all is kept as a passthrough rather than removed, since dropping the target would change the CMake surface.

Type of change: refactor / cleanup

How to verify: clean configure and build with cmake -B build -D SIMDJSON_DEVELOPER_MODE=ON; the six packages no longer appear in the CPM output. ctest --test-dir build passes 133/133, unchanged. The trimmed checker compiles standalone (g++ -std=c++17 -c with rapidjson and sajson includes); the full allparserscheckfile target is not built, due to the guard described above. No performance claims.

Fixes #2859

---
AI disclosure: prepared with AI assistance; all diffs reviewed, compiled, and tested by author.


lemire:
Running tests.

PR #2863, opened 2026-09-17
  perf(arm64): optimize Stage 1 operator matching

mpurbay-arm:
## Description
Use SVE2 MATCH for JSON operator classification when SVE2 and the
NEON-SVE bridge are available. This replaces the existing NEON
table-lookup sequence with direct operator matching.

Convert the four resulting SVE predicates into the compact 64-bit
operator mask required by Stage 1.

Retain the existing operator-classification path when the SVE2 path
is unavailable. Other architectures, including x86, are unaffected.

Issue reproduced / related issue: N/A

## Type of change
- [ ] Bug fix
- [x] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

## How to verify / test
Validated Arm64 builds with GCC and Clang. Performance was evaluated
across the 22 standard simdjson benchmark inputs.

Additional checks:
- Verified the SVE2 MATCH and NEON-SVE bridge path.
- Confirmed that non-Arm implementation files remain unchanged.
- Ran the complete test suite with AddressSanitizer and 
UndefinedBehaviorSanitizer using Clang; all tests passed without
sanitizer reports.

## Checklist before submitting
- [ ] I added/updated tests covering my change (existing tests cover the changed paths)
- [x] Code builds locally and passes my checks
- [x] Documentation / README updated if needed (not required)
- [ ] I linked the related issue (not applicable)


lemire:
@mpurbay-arm Can you help me address the following concern ?


SVE is vector length agnostic. But, we are using the registers as if they were 128-bit registers. Clearly, on hardware where the registers are much larger (say 1024 bits), it would be silly code.

How do we handle that?

mpurbay-arm:
Hi @lemire 
Thanks for raising this, you are right. 

- This is not a fully vector-length-agnostic use of SVE2. The existing Stage 1 implementation processes each 64-byte block as four 16-byte NEON chunks. This change preserves that structure and uses SVE2 MATCH only on the lower 16 byte lanes of each chunk.

- Correctness does not depend on the hardware SVE vector length. svptrue_pat_b8(SV_VL16) activates only the lowest 16 byte lanes, so upper lanes cannot contribute to the match result. The predicate-store buffers are also sized for the architectural maximum SVE vector length, making wider predicate stores safe.

- Therefore, the implementation remains correct on wider-VL systems. The limitation is that it continues to process 16 bytes per MATCH and does not use the additional lanes.

- Performance has currently been validated with a 128-bit SVE vector length. This patch optimizes the existing four 16-byte Stage 1 chunks; it does not introduce the 128-bit chunking limitation.

lemire:
@mpurbay-arm Yes, I understand that, but if someone comes up with a 1024-bit SVE implementation, then this implementation here will do compute on 1024-bit registers but keep only the low 128-bit register. It seems quite wasteful.

It is good that, right now, most SVE2 implementations are just using 128-bit registers, but should this expand in the future (to, say, 512-bit) then this code here would become wasteful, wouldn't it?

clausecker:
I have no experience with that, but I presume long SVE implementations will behave like good vector implementations in that instructions masked to shorter lengths execute with a lower latency.

lemire:
@clausecker I'll go with this assumption.

lemire:
@mpurbay-arm  I think we will use the SVE2 optimization. I have covered it on my blog in the past, but never tested it. Your PR shows that it can be beneficial, at least on Neoverse V cores.

The other optimization you introduced is separate and it is CPU specific. Extracting bits out of a word is expensive in ARM, but an Apple CPU has plenty of slack and is mostly bounded by data dependencies. ARM cores are weaker and benefit from your optimization. I think it is worth investigating further, but as a separate issue from SVE2.

mpurbay-arm:
@lemire 
Thanks, that makes sense. 

- I’ll keep this PR focused on the SVE2 MATCH classifier and remove the independent Mixed4 writer change. 
- I’ll investigate the writer optimization separately with broader Arm CPU validation before proposing it in a separate PR.

src/arm64.cpp
```diff
@@ -37,10 +75,48 @@ namespace {
...
+    ',', ':', '[', ']', '{', '}', ',', ':'
+  };
+  const svbool_t pg = svptrue_pat_b8(SV_VL16);
+  const svuint8_t data = svset_neonq_u8(svundef_u8(), input);
+  const svuint8_t table = svset_neonq_u8(svundef_u8(), operators);
+  return svmatch_u8(pg, data, table);
+}
+
+#define SIMDJSON_ARM64_STORE_PREDICATE(PREDICATE, DESTINATION) \
+  __asm__ volatile("str %0, [%1]" : : "Upl"(PREDICATE), "r"(DESTINATION) : "memory")
```
clausecker, line 53: Why is this inline assembly needed?  I don't quite see the point.

mpurbay-arm: @clausecker Thanks for asking about this.

svmatch_u8 returns an svbool_t, while Stage 1 needs a compact mask with one bit per input byte. STR P stores a predicate register in this packed-bit form. We store four predicates, read the first 16 bits from each, and combine them into one uint64_t.

Inline assembly is used because ACLE does not provide a direct intrinsic for STR P. svst1 is different: it stores vector data under predicate control rather than storing the predicate itself.
 
I tested alternatives including fixed-length predicate types, a packed single-load implementation, and a register-only conversion. They generated more instructions or performed worse. With the tested GCC and Clang versions, the current implementation generates four STR P stores followed by four LDRH loads, without a memcpy function call.

I will add a short source comment explaining why STR P is used here.

clausecker: There is no intrinsic because you can just use the `=` operator to store the predicate to memory.  Then do some type-punning and you should be good to go.  If you want this to be strict aliasing compliant, store into a separate buffer and then use memcpy to transfer the contents.

Make sure the array is long enough—storing a predicate stores a CPU-dependent number of bytes, up to a maximum of 2048/8 = 256 bytes.  Your array is not 256 bytes long.

mpurbay-arm: @clausecker Thanks. I tested the assignment approach using a fixed-length predicate type:

```cpp
using predicate128 =
    svbool_t __attribute__((arm_sve_vector_bits(128)));

const predicate128 predicates[4] = {p0, p1, p2, p3};

uint64_t result;
__builtin_memcpy(&result, predicates, sizeof(result));
```

This requires `-msve-vector-bits=128`, because scalable `svbool_t` is sizeless and cannot be used as an array element. It passed the complete test suite with both GCC and Clang.

However, with the compiler versions I tested, both generated extra predicate spills, reloads and stack operations compared with the explicit `STR P` implementation. It also performed worse across every Stage 1 benchmark input I tested. The explicit `STR P` version generated simpler code.

I also created a reduced Compiler Explorer example using Clang 22.1 with:

```text
-O3 -march=armv9-a+sve2 -msve-vector-bits=128
```

[https://godbolt.org/z/KzEaK3rqo](https://godbolt.org/z/KzEaK3rqo)

In this example, the explicit version generates four `STR P` stores followed by four `LDRH` loads, while the fixed-length assignment version adds extra predicate reloads, lane moves and stack operations. I observed the same general behavior with GCC 16.1.

Regarding the buffer size, an SVE predicate has one bit for each vector byte. At the maximum 2048-bit SVE vector length there are 256 byte lanes, so the predicate contains 256 bits, which is 32 bytes. Therefore, each existing 32-byte buffer is large enough.

lemire:
@mpurbay-arm  I moved this to intrinsics only.  See https://github.com/simdjson/simdjson/pull/2866

It is less efficient than your approach, but I don't want to rely on luck at this time. This code can end up in production, on systems that I cannot test.

mpurbay-arm:
Thanks @lemire  I reviewed #2866. The svsel_u8 conversion followed by the existing pairwise-add reduction is a clean intrinsics-only solution. It avoids inline assembly, fixed-length predicate types, and predicate memory-layout concerns.

I understand the trade-off: it is slightly less efficient than the STR P implementation, but it is safer for production systems and different compiler/SVE configurations.

I am happy for #2866 to supersede #2863. Thank you for retaining the attribution.

mpurbay-arm:
@lemire 
Thanks also for splitting the Mixed4 writer into #2869 and validating it independently across a broader range of Arm64 CPUs.

Together, #2866 and #2869 cover the two optimizations from this PR while keeping the classifier and writer changes separate and easier to review.

Thank you for retaining the attribution and taking this work forward.

src/arm64.cpp
clausecker: I'm not sure why you think that you need an array of `svbool_t`.  What I suggest is something like

    uint8_t bits[...];

    *(*svbool_t)&bits[0] = predicate0;
    *(*svbool_t)&bits[1] = predicate1;

and so on, such that the predictes are stored with a one byte overlap in `bits`.  This requires that strict aliasing be turned off for the function, which you can do with some attribute I believe.  Note that the `bits` array needs to be long enough to store the maximum length of an `svbool_t` (2048 bits) at its maximum offset.  I.e. add 255 to the length to account for this.

PR #2866, opened 2026-09-18
arm64: classify structural characters with SVE2 MATCH when available

lemire:
This is a continuation of https://github.com/simdjson/simdjson/pull/2863 taken from @mpurbay-arm. The main difference between https://github.com/simdjson/simdjson/pull/2863 and this PR here is that we avoid inline assembly and undocumented approaches. This cost us a bit of performance in practice, but should open us to fewer bad surprises.

One bad surprise that we cannot rule out is that people will run this on processors with wide SVE registers (say 512 bits) and waste 512-128 bits of computation. For now, I am accepting @clausecker's stance that, hopefully, these vector ISA might someone realize that we are just using the low 128 bits and not waste too much compute. I am not sure how reliable this assumption is, but I choose to trust it for now.

Importantly, this whole thing only works when the compiler targets SVE2 with the NEON-SVE bridge (e.g. `-mcpu=native` on Graviton 4/5, or `-march=armv9-a+sve2`). I will open an issue about supporting runtime dispatching for SVE2.

So, when SVE2 is available replace the NEON table-lookup sequence that identifies the structural characters (, : [ ] { }) in stage 1 with a single SVE2 MATCH per 16-byte chunk. Whitespace is still classified with the NEON lookup.

The four MATCH predicates are turned back into a 64-bit mask without going through memory: a predicated select materializes each predicate as bit weights, and the usual pairwise-add tree folds them. The last pairwise add is now shared between the operator and whitespace trees on all arm64 builds (operators land in lane 0, whitespace in lane 1), saving one instruction on the NEON path as well.

Only the low 16 lanes are active (SV_VL16) and only the low 128 bits of the SVE registers are read, so the code is correct for any SVE vector length; it was checked under QEMU at 128, 256 and 512 bits with GCC 16 and clang 21. 

Performance: `benchmark/dom/parse -t -n 300` over the 22 simdjson-data files, `-mcpu=native`, Release, static, pinned to one core, best of 3 interleaved runs (noise < 1%). Stage-1 throughput ratio (SVE2 / NEON), geometric mean over the files:

|     | Graviton 4 (Neoverse V2) | Graviton 5 (Neoverse V3) |
| --- | --- | --- |
| GCC 15.2 | 1.06 (5.45 → 5.78 GB/s) | 1.03 (7.09 → 7.29 GB/s) |
| clang 21.1 | 1.09 (4.83 → 5.28 GB/s) | 1.04 (6.28 → 6.55 GB/s) |

Whole-parse: +1% to +4%. Stage 2 unchanged. Largest per-file gains on `gsoc-2018` (+9% to +16%) and `random` (+5% to +14%); only `canada` on Graviton 5/GCC is marginally down (−3%).

So it is not pure gain, some cases might be slower.

Credit is given to @mpurbay-arm in the commit.


lemire:
I might hold on before merging this that we have runtime dispatching.

lemire:
Working on runtime detection at https://github.com/simdjson/simdjson/pull/2867

PR #2867, opened 2026-09-18
Arm64 sve runtime detection

lemire:
For features such as https://github.com/simdjson/simdjson/pull/2866  I would prefer to rely on runtime dispatching. This PR is the first step in this direction.


src/internal/isadetection.h
```diff
@@ -74,8 +82,57 @@ static inline uint32_t detect_supported_architectures() {
...
+  // NEON is mandatory on AArch64.
+  uint32_t host_isa = instruction_set::NEON;
+#if defined(__linux__)
+  unsigned long hwcap = getauxval(AT_HWCAP);
+  unsigned long hwcap2 = getauxval(AT_HWCAP2);
+  if (hwcap & HWCAP_SVE) {
+    host_isa |= instruction_set::SVE;
+  }
+  if (hwcap2 & HWCAP2_SVE2) {
+    // SVE2 implies SVE.
```
mediouni-m: There's a bit of a bad surprise around this on older Linux kernels where this was not true.

`sve2` was advertised on processors that implemented SME(2) but not SVE until https://lore.kernel.org/r/20250107-arm64-2024-dpisa-v5-1-7578da51fc3d@kernel.org (Linux 6.14 onwards). Because SVE2 instructions were exposed in streaming mode.

For use of SVE instructions, always checking for HWCAP_SVE is safe.

mediouni-m (changes requested):

src/internal/isadetection.h
lemire: @mediouni-m Can you check again ? My goal is to start using SVE2 more aggressively.

mediouni-m: Yep now looks fine :)

PR #2870, opened 2026-09-19
Fix line break handling for fractured json formatting

maxbachmann:
This brings formatting a lot closer to https://github.com/j-brooke/FracturedJson/wiki by fixing a bunch of issues with the handling of the line break calculation. 

Type of change
- [x] Bug fix
- [ ] Optimization
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation / tests

There is another big topic this doesn't touch yet which is table formatting. This has severe issues but I wasn't sure whether we essentially want to handle them close to the original Fractured Json:
- it doesn't use line lengths to decide whether it should use table formatting
- it reorders the elements in an alphabetic way for the matching keys. This happens even when the keys already have a consistent ordering. The original orders by first occurrence. When a key doesn't exist in a line it adds spacing in between to skip the position.
- it has no separate complexity threshold similar to the other parts (It has a similarity threshold the original doesn't have though)

For me this handling in the original FracturedJson works quite well, but I wanted to make sure this is what we want here before making the change since it's a slightly larger change.


maxbachmann:
Also I wasn't sure if every commit should update the amalgation via the script or when this is done

lemire:
> Also I wasn't sure if every commit should update the amalgation via the script or when this is done

No.  We refresh on release or periodically, but not on every commit.

lemire:
Running tests.

maxbachmann:
@lemire what about table formatting? Is the goal to behave similar to Fractured JSON or what is the goal?

lemire:
@maxbachmann Yes, we should be Fractured JSON compliant ? Whatever this might mean?

PR #2874, opened 2026-09-22
builder: restore `pos + n <= cap` in writer::ensure (citm serialization +14-20%)

lemire:
## Summary

Commit a16e5128 ("pendantic guard", May 7) rewrote the hot capacity check in `writer::ensure()` from `pos + n <= cap` to `n <= cap - pos`, to rule out a wrap of `pos + n` on 32-bit targets. That one-line change turns out to cost **14-20% of serialization throughput on `citm_catalog.json`** under GCC 16.

This PR restores `pos + n <= cap` and moves the wrap protection to the only place where `n` is data-dependent: the string and map-key atoms now check `size() > (SIZE_MAX - 2 - pos) / 6` (resp. `- 3 - pos`), which proves `pos + 2 + 6*size` cannot wrap before calling `ensure()`. Every other `ensure()` caller passes a `constexpr` key/enum length. The safety argument of a16e5128 is therefore fully preserved.

## Why it was slow

`cap - pos` is a "remaining" temporary that has to stay live alongside `pos` across the fully inlined serializer, whereas `pos + n` is the value the following write needs anyway. GCC cannot canonicalize between the two forms because they differ under wrap. The inlined citm `append()` grew from 129 KB to 145 KB with ~11% more loads/stores (register pressure / spills). `citm_catalog` (integer- and key-heavy, many back-to-back `ensure(20)`/`ensure(1)`) is hit; `twitter` (dominated by long strings inside `write_string_escaped`) is not.

## Measurements

GCC 16.1 (arm64), `-O3 -freflection`, static reflection; four binaries built from one header via `-D` toggles, run interleaved, max-of-best MB/s:

| bench | before | `ensure` reverted only | size guards dropped only | this PR |
|---|---|---|---|---|
| citm/reuse_buffer | 9058 | +20.6% | −0.3% | **+20.0%** |
| citm/static_reflection | 8182 | +13.9% | +4.2% | **+15.8%** |
| citm/to | 7304 | +14.5% | −0.2% | **+13.8%** |
| citm/to_reuse | 7318 | +14.4% | −0.4% | **+14.1%** |
| twitter (all four) | — | ±2% | ±2% | ±2% (noise) |

The extra guard branches are free; the whole cost was the `ensure()` rewrite, and this PR recovers all of it.

Only measured on GCC 16 / arm64 so far; given the mechanism I expect the same direction elsewhere but have not verified.

## Test plan

- [x] `tests/builder` (8 tests) rebuilt against the new header under GCC 16 and pass
- [ ] CI


include/simdjson/generic/builder/json_builder.h
```diff
@@ -147,11 +149,11 @@ simdjson_really_inline constexpr void atom(writer &w, const T &t) {
...
   }
   // Worst-case escape: every byte expands to \uXXXX (6 chars), plus 2 quotes.
-  // Guard against 2 + 6 * input.size() wrapping for huge inputs -- if it
-  // wrapped to a small value, ensure() would spuriously succeed and the
+  // Guard against w.pos + 2 + 6 * input.size() wrapping for huge inputs -- if
+  // it wrapped to a small value, ensure() would spuriously succeed and the
   // subsequent escape would overflow the buffer.
   // Note that this is pedantic except maybe on 32-bit targets.
-  if (simdjson_unlikely(input.size() > ((std::numeric_limits<size_t>::max)() - 2) / 6)) { return; }
+  if (simdjson_unlikely(input.size() > ((std::numeric_limits<size_t>::max)() - 2 - w.pos) / 6)) { return; }
```
lemire: **[bug]** `(SIZE_MAX - 2 - w.pos) / 6` is evaluated in `size_t`. When `w.pos > SIZE_MAX - 2` (`pos >= SIZE_MAX - 1`), the subtraction wraps to a value near `SIZE_MAX`, the quotient stays huge, and the branch does not fire — including for `size == 0`, where `n` is only 2. `ensure` then adds `pos + n`, which wraps. Any wrapped sum is strictly less than `pos`, and `pos <= cap`, so `pos + n <= cap` is true and `grow_slow`'s `pos + n < pos` test is skipped. The quote and `write_string_escaped` stores that follow run past the buffer. For every `pos <= SIZE_MAX - 2` the predicate is sufficient (`6 * size` cannot wrap either). The comment immediately above this line says the guard exists so a wrapped length cannot make `ensure` succeed; that is false for the top two `pos` values, which are exactly the 32-bit case the guard claims to cover, and `ensure` no longer backs it up.

**Suggestion:** Reject before subtracting, e.g. `w.pos > (std::numeric_limits<size_t>::max)() - 2 || input.size() > ((std::numeric_limits<size_t>::max)() - 2 - w.pos) / 6`. Keep this off the `ensure` hot path.

include/simdjson/generic/builder/json_builder.h
```diff
@@ -178,11 +180,11 @@ simdjson_really_inline constexpr void atom(writer &w, const T &m) {
...
     // Keys must be convertible to string_view per the concept.
     std::string_view key_sv(key);
-    // Guard against 3 + 6 * key_sv.size() wrapping for huge keys, if it
-    // wrapped to a small value, ensure() would spuriously succeed and the
+    // Guard against w.pos + 3 + 6 * key_sv.size() wrapping for huge keys, if
+    // it wrapped to a small value, ensure() would spuriously succeed and the
     // subsequent escape would overflow the buffer.
     // Note that this is pedantic except maybe on 32-bit targets.
-    if (simdjson_unlikely(key_sv.size() > ((std::numeric_limits<size_t>::max)() - 3) / 6)) { return; }
+    if (simdjson_unlikely(key_sv.size() > ((std::numeric_limits<size_t>::max)() - 3 - w.pos) / 6)) { return; }
```
lemire: **[bug]** Same unsigned subtraction hole as the string atom, with constant 3. `(SIZE_MAX - 3 - w.pos) / 6` wraps when `w.pos > SIZE_MAX - 3` (`pos >= SIZE_MAX - 2`). An empty key (`size == 0`, `n == 3`) is not rejected, `pos + 3` wraps, and `ensure(2 + 6 * key_sv.size() + 1)` takes the fast path. The following quote, escape, and `:` stores overflow. For every `pos <= SIZE_MAX - 3` the check does match `3 + 6 * size` (the `ensure` argument is `2 + 6 * size + 1`).

**Suggestion:** Reject first with `w.pos > (std::numeric_limits<size_t>::max)() - 3 || key_sv.size() > ((std::numeric_limits<size_t>::max)() - 3 - w.pos) / 6`.

include/simdjson/generic/builder/json_builder.h
```diff
@@ -68,10 +68,12 @@ struct writer {
...
-    // Use subtraction (relying on the pos <= cap invariant) so a huge n
-    // cannot wrap pos + n to a small value that spuriously passes the test.
-    // This is pedantic except maybe on 32-bit targets.
-    if (simdjson_likely(n <= cap - pos)) return true;
+    // Callers guarantee that pos + n does not wrap: n is either a small
+    // constant or has been checked against pos (see the string atoms).
+    // Do not rewrite this as `n <= cap - pos`: that form keeps an extra
+    // live value across the fully inlined serializer and costs GCC 16
+    // 14-20% of throughput on citm_catalog.json.
+    if (simdjson_likely(pos + n <= cap)) { return true; }
```
lemire: **[suggestion]** Lines 71–72 say callers guarantee `pos + n` does not wrap because `n` is a small constant or was checked against `pos`. Only the two string/map atoms inspect `pos`. Every other call in this header passes a literal or a constexpr length: `ensure(1)`, `ensure(2)`, `ensure(4)`, `ensure(5)`, `ensure(20)` (booleans and both integer arms), and `ensure(rest_key_len)` / `ensure(enum_str_len)` in the struct, enum, and `extract_from` atoms. For those, `pos + n` wraps whenever `pos > SIZE_MAX - n` (`pos == SIZE_MAX` is enough for `ensure(1)`; `pos >= SIZE_MAX - 19` for the integer `ensure(20)`). The wrapped sum compares `<= cap` and is treated as already reserved, so the following `memcpy`, single-byte store, or `write_uint_jeaiii` writes off the end. `grow_slow` cannot catch it: its `pos + n < pos` test is unreachable from `ensure` once the fast path accepts every wrapped sum. A constexpr length is not a proof; it only shrinks the window to `SIZE_MAX - rest_key_len`. Leaving literals unchecked is a fair tradeoff — hitting it requires a successful allocation larger than `SIZE_MAX - n`, which a 32-bit process cannot place — but a16e512's "safe for every `n`" argument is not fully preserved, and the comment should not say callers guarantee it.

**Suggestion:** Keep `pos + n <= cap`. State the real precondition (`n <= SIZE_MAX - pos`) instead of claiming every caller establishes it. Do not add a `pos > SIZE_MAX - n` branch on the literal hot path unless a 32-bit measurement shows it is free.

include/simdjson/generic/builder/json_builder.h
```diff
@@ -68,10 +68,12 @@ struct writer {
...
   simdjson_really_inline bool ensure(size_t n) noexcept {
-    // Use subtraction (relying on the pos <= cap invariant) so a huge n
-    // cannot wrap pos + n to a small value that spuriously passes the test.
-    // This is pedantic except maybe on 32-bit targets.
-    if (simdjson_likely(n <= cap - pos)) return true;
+    // Callers guarantee that pos + n does not wrap: n is either a small
+    // constant or has been checked against pos (see the string atoms).
+    // Do not rewrite this as `n <= cap - pos`: that form keeps an extra
+    // live value across the fully inlined serializer and costs GCC 16
+    // 14-20% of throughput on citm_catalog.json.
```
lemire: **[suggestion]** The new comment on lines 71–75 is a change log: it names the rejected `n <= cap - pos` form, GCC 16, a 14–20% `citm_catalog.json` delta, and the inlined serializer's live values. That history belongs in the commit message. The durable constraint is only why this expression must stay an addition.

**Suggestion:** Replace it with one line, for example that `n <= cap - pos` is an extra live value across the inlined serializer and regresses throughput. Drop the benchmark numbers and the compiler version.

